PR #5, opened 2025-02-21
Changed skipsdist to False in tox.ini to enable source distribution

phantomii:


.github/workflows/tests.yml
```diff
@@ -19,6 +19,8 @@ jobs:
           python-version: ${{ matrix.python-version }}
       - name: Install tox
         run: sudo apt update && sudo apt install --yes tox
+      - name: Install EV library
+        run: sudo apt update && sudo apt install --yes libev-dev
```
akremenetsky, line 23: Why do we want to install `libev-dev` as a separate step? We can merge these two step into one something like:
```yaml
- name: Install packges
  run: sudo apt update && sudo apt install --yes tox libev-dev
```

akremenetsky, line 23:
```suggestion
      - name: Install requirements
        run: sudo apt update && sudo apt install --yes tox libev-dev
```

PR #6, opened 2025-02-22
Add init migration structure

phantomii:


migrations/0001-init-all-tables-f47bad.py
```diff
@@ -0,0 +1,48 @@
...
+#    under the License.
+
+from restalchemy.storage.sql import migrations
+
+
+class MigrationStep(migrations.AbstarctMigrationStep):
+
+    def __init__(self):
+        self._depends = [
+            "0000-init-iam-tables-9c9153.py",
```
gmelikov, line 25: Do we really want to use tree-like dependencies? I thought we usually use linear dependencies, in this case there should be one 0000 migration.

phantomii: I want us to be able to conveniently debug logically separated code segments at this stage. Reloading the database of a specific logical module is useful during development. For example, if I’m working on IAM functionality, I can just run `ra-rollback -m iam && ra-apply -m iam`. The rest of the database remains untouched, and I won’t have to manually push data into artifacts or nodes. Additionally, splitting the project into separate repositories when needed would make things easier to manage.

gmelikov (approved):

akremenetsky (approved):

PR #34, opened 2025-03-23
Filters and waighters for scheduler

akremenetsky:
The scheduler logic was reworked to use filters and weighers. The first step is to filter our all unsuitable machine pools. The second step is to chose the best machine pool based on various parameters. For instance choose the most free machine pool and so on.

### DIFF
--- a/genesis_core/node/scheduler/driver/weighter/relative.py
+++ b/genesis_core/node/scheduler/driver/weighter/relative.py
@@ -1,0 +1,66 @@
+#    Copyright 2025 Genesis Corporation.
+#
+#    All Rights Reserved.
+#
+#    Licensed under the Apache License, Version 2.0 (the "License"); you may
+#    not use this file except in compliance with the License. You may obtain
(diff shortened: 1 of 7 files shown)


genesis_core/node/scheduler/driver/weighter/relative.py
```diff
@@ -0,0 +1,66 @@
...
+        Every machine pool gets a weight from range [0, 1].
+        1 means the pool is the best for the node.
+        0 means the pool is the worst for the node.
+        """
+
+        # Maximum weight gets a pool with relative maximal cores and ram
+        usages = tuple(self._usage_ratio(p) for p in pools)
+
+        # The system is empty, all pools have equal weight
+        if sum(usages) == 0:
```
gmelikov, line 63: NIT for future: we should SUM once, and reuse later if needed.

gmelikov (approved):

phantomii (approved):

PR #36, opened 2025-03-24
Allow unannotated fields for Izulu exceptions

akremenetsky:
The default behaviour of Izulu package was changend, it restricted unannotated fields. Enable original behaviour using special options.


gmelikov (approved):

gmelikov:
filEds, changeNd

phantomii (approved):

PR #39, opened 2025-03-28
Add notification to confirmation email

phantomii:


etc/genesis_core/event_type_mapping.yaml
```diff
@@ -0,0 +1,3 @@
+EventTypeMapping:
+    IamUserRegistration: b963bcfe-0a99-11f0-916d-047c160cda6f  # genesis_core/events/constants.py:IAM_USER_REGISTRATION_EVENT
```
akremenetsky, line 2: What is this? These UUIDs will be default for every Genesis installations?

genesis_core/cmd/user_api.py
```diff
@@ -119,19 +127,31 @@ def main():
...
+            )
+        )
+
+        serv_hub.add_service(service)
 
-    log.info("Bye!!!")
+    if CONF[DOMAIN].workers > 1:
+        serv_hub.start()
+    else:
+        service.start()
```
akremenetsky, line 154: Why we would like to use service? Can we use only `hub` here?

genesis_core/common/utils.py
```diff
@@ -48,6 +49,7 @@ def load_group_from_entry_point(group: str) -> tp.Any:
 def get_context_storage(
     global_salt: str,
     token_algorithm: algorithms.AbstractAlgorithm,
+    events_client: sdk_clients.AsyncEventClient,
```
akremenetsky, line 52: Does only an `async` client suit here? Can we use any other client, let's say http client?

genesis_core/cmd/user_api.py
phantomii: Using a hub causes debugging problems. VS Code breakpoints aren’t hit.

etc/genesis_core/event_type_mapping.yaml
phantomii: needs to be configured for each installation, but the element manager subsystem does not exist yet, so hardcode.

akremenetsky (approved):

etc/genesis_core/event_type_mapping.yaml
```diff
@@ -0,0 +1,3 @@
...
+    ResetPassword: "df5ac4b0-de7e-41d5-a0bc-6e26bf35594a"
```
gmelikov, line 3: Wrong? Something like IamUserResetPassword

PR #47, opened 2025-04-27
Base implementation of baremetal node support

akremenetsky:
Main focus in this change is the basic baremetal support. But also a log
of other many improvements are introduced. Main of them are:

- First implementation of Status API for Core agents. Core agents is
  able to register thier `payloads` and actualize them during life
  cycle.
- Default HW machine pool. All new baremetal nodes are added to this
  pool after if the follow the auto discovery procedure.
- Auto discovery IP range. To support the auto discovery procedure for
  baremetal nodes.
- Reworked machine scheduling logic. Scheduler filters and weighers for machines
  were added.

### DIFF
--- a/genesis_core/network/dm/models.py
+++ b/genesis_core/network/dm/models.py
@@ -20,15 +20,21 @@ from restalchemy.dm import relationships
 from genesis_core.node.dm import models
 
 
-class Port(models.Port, models.CastToBaseMixin):
-    __cast_filels__ = ("node", "machine", "subnet")
+class Subnet(models.Subnet, models.CastToBaseMixin):
+    __cast_fields__ = ("network",)
 
-    subnet = relationships.relationship(models.Subnet, prefetch=True)
+    network = relationships.relationship(models.Network, prefetch=True)
+
+
+class Port(models.Port, models.CastToBaseMixin):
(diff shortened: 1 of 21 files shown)


genesis_core/network/dm/models.py
```diff
@@ -20,15 +20,21 @@
 from genesis_core.node.dm import models
 
 
+class Subnet(models.Subnet, models.CastToBaseMixin):
+    __cast_filels__ = ("network",)
```
gmelikov, line 24: filels?

genesis_core/node/dm/models.py
```diff
@@ -503,3 +543,64 @@ class NodeWithoutPorts(Node):
...
+    def get_vm_nodes(cls):
+        return cls.objects.get_all(
+            filters={
+                "node_type": dm_filters.EQ(nc.NodeType.VM.value),
+            }
+        )
+
+
+class HWNodeWithoutPorts(models.ModelWithUUID, orm.SQLStorableMixin):
+    __tablename__ = "compute_hw_nodes_without_ports"
```
gmelikov, line 557: Do we really need explicit `_without_ports` suffix?

akremenetsky:
Reworked and closed in https://github.com/infraguys/genesis_core/pull/60

PR #51, opened 2025-04-29
Initial ZFS support

gmelikov:


genesis_core/node/machine/pool/driver/libvirt.py
```diff
@@ -484,8 +510,20 @@ def delete_volume(self, volume: models.MachineVolume) -> None:
...
                 if v.name().startswith(str(volume.uuid)):
-                    v.wipe()
-                    v.delete()
+                    try:
+                        v.wipe()
+                    except libvirt.libvirtError as e:
+                        # Some backends don't need wiping, for ex. ZFS
+                        if e.get_error_code() != 3:  # VIR_ERR_NO_SUPPORT
+                            raise
+                    for i in range(20):
```
akremenetsky, line 519: Should we add `else` part for the `for` in case if the volume won't be delete? In this part we may raise an exception to indicate that something goes wrong and see it in logs.

akremenetsky (approved):

phantomii (approved):

PR #56, opened 2025-05-13
Config service and Core agent

akremenetsky:
Config service and Core agent
    
This change adds two significant improvements:

- Config service
The service that allows to manager life cycle of configuration files.
The first implementation support plain text and a single node as a
target where the configuration will be delivered. Also you may specify
permission mode, group and owner for the configuration. An example of
config creation may look like this:

```json
{
    "name": "test-config",
    "project_id": "c27f59fe-d3d3-4625-804c-5c9fd3bb3142",
    "path": "/etc/my_config.conf",
    "target": {
        "kind": "node",
        "node": "0712d955-4f59-4633-862e-54e943b232f4"
    },
    "body": {
        "kind": "text",
        "content": "My first config!\n"
    }
}
```

The configuration will be placed to the `0712d955...` node by
`/etc/my_config.conf` path. Also it's possible to specify additional
actions that has to be executed if the configuration is changed.

- Core agent
The second big part of this change is introducing of the core agent.
This agent will be placed into base images and it will be act as
feedback for the orchestrator. Also it will perform actual actions on
the data plance - create configs, start services, update networks and so
on.

### DIFF
--- a/etc/genesis_core_agent/genesis_core_agent.conf
+++ b/etc/genesis_core_agent/genesis_core_agent.conf
@@ -1,0 +1,6 @@
+[DEFAULT]
+verbose = True
+debug = True
+
+[core_agent]
+orch_endpoint = http://10.20.0.2:11011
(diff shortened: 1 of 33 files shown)


akremenetsky:
Reworked and closed in https://github.com/infraguys/genesis_core/pull/60

PR #60, opened 2025-05-30
Universal agent

akremenetsky:
Because of historical reasons three topic in this PR.

### Base implementation of baremetal node support
Main focus in this change is the basic baremetal support. But also a log
of other many improvements are introduced. Main of them are:

- First implementation of Status API for Core agents. Core agents is
able to register thier payloads and actualize them during life
cycle.
- Default HW machine pool. All new baremetal nodes are added to this
pool after if the follow the auto discovery procedure.
Auto discovery IP range. To support the auto discovery procedure for
baremetal nodes.
- Reworked machine scheduling logic. Scheduler filters and weighers for machines
were added.

### Config service

The service that allows to manager life cycle of configuration files.
The first implementation support plain text and a single node as a
target where the configuration will be delivered. Also you may specify
permission mode, group and owner for the configuration. An example of
config creation may look like this:
```
{
    "name": "test-config",
    "project_id": "c27f59fe-d3d3-4625-804c-5c9fd3bb3142",
    "path": "/etc/my_config.conf",
    "target": {
        "kind": "node",
        "node": "0712d955-4f59-4633-862e-54e943b232f4"
    },
    "body": {
        "kind": "text",
        "content": "My first config!\n"
    }
}
```
The configuration will be placed to the `0712d955...` node by
/etc/my_config.conf path. Also it's possible to specify additional
actions that have to be executed if the configuration is changed.

### Universal agent adaptation
- Universal agent services added to the genesis core image.
- Config service was reworked to use entities from the universal agent.


genesis_core/node/scheduler/service.py
```diff
@@ -169,6 +185,92 @@ def _schedule_nodes(self) -> tp.List[models.Machine]:
...
+                    "The node %s scheduled to %s machine",
+                    node.uuid,
+                    machine.uuid,
+                )
+            except Exception:
+                LOG.exception("Error scheduling node %s", node.uuid)
+
+            # Actualize idle machines
+            if node.node_type == nc.NodeType.HW.value:
+                idle_hws.remove(machine)
```
gmelikov, line 268: idle_hws/idle_vms is not returned to anywhere, and it's used only as `idle_machines = idle_hws`. Maybe I miss something, but looks like we can just drop these removals, no?

genesis_core/orch_api/dm/models.py
```diff
@@ -126,49 +135,71 @@ def from_agent_payload(
...
-
-        if interfaces := payload.get("interfaces"):
-            data["interfaces"] = [
-                {
-                    "mac": iface["mac"],
-                    "ipv4": iface["ipv4"],
-                    "mask": iface["mask"],
-                }
-                for iface in interfaces
+        payload = copy.deepcopy(payload)
```
gmelikov: Looks like it's a hot path, can we exclude this deepcopy somehow?

migrations/0017-core-agent-4b8026.py
```diff
@@ -76,7 +76,7 @@ def upgrade(self, session):
             """,
             """
             CREATE UNIQUE INDEX IF NOT EXISTS compute_net_interfaces_mac_machine_id_idx
-                ON compute_ports (mac, machine);
+                ON compute_net_interfaces (mac, machine);
```
gmelikov: Is it really ok to edit already existing migration?

akremenetsky: We don't have this snippet in the final version of PR. Commits will be squashed.

genesis_core/config/service.py
```diff
@@ -97,32 +113,41 @@ def _actualize_new_config(
...
             return
 
+        config_resource = config.to_ua_resource(CONFIG_CAPABILITY_RESOURCE)
+        config_resource.insert()
+
         # Make renders for this config
         for node in target_nodes:
             render = config.render(node=node.uuid)
 
+            # Hach for scheduler
```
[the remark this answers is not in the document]
akremenetsky: Fixed. Sorry not accepting your suggestion I was in the middle of the squashing procedure.

There are three commits now in according to description of this PR.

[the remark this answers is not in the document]
gmelikov: That's ok, suggestion is an easy way to show proposition via diff, I personally prefer not to add them as commits, but interpret as ideas.

phantomii (approved):

gmelikov (approved):

PR #61, opened 2025-05-30
Introduce basic DNSaaS

gmelikov:


genesis/images/install.sh
```diff
@@ -95,3 +95,14 @@ sudo cp "$GC_PATH/etc/systemd/gc-gservice.service" $SYSTEMD_SERVICE_DIR
 
 # Enable genesis core services
 sudo systemctl enable gc-user-api gc-orch-api gc-gservice
+
+
+# Prepare DNSaaS
+
+# Install packages
+sudo apt install pdns-backend-pgsql pdns-server -y
```
akremenetsky, line 103: Can we merge package installation into single command? I mean with the command above. It will be a bit faster.

gmelikov: While I prefer speed too, for now it's much easier to track dataplane preparations separately, until further extensive production experience, I propose to leave it as is.

akremenetsky: Ok

akremenetsky (approved):

phantomii (approved):

PR #76, opened 2025-07-04
Dns: introduce dnsdist as resolver, use it for all VMs

gmelikov:


genesis_core/cmd/bootstrap.py
```diff
@@ -77,6 +80,8 @@ def _apply_startup_db():
             ports = subnet.pop("ports", [])
             subnet_project_id = subnet.pop("project_id", net_project_id)
             subnet["project_id"] = subnet_project_id
+            if startup_entities.get("domain"):
```
akremenetsky, line 83: `if startup_entities.get("core_ip"):` ?

phantomii (approved):

genesis_core/cmd/bootstrap.py
gmelikov: I made it on purpose, core should have some ip, but we may not want to use pdns==local domain (for now).

I think in future we'll remove such conditions at all.

akremenetsky: Got it.

akremenetsky (approved):

PR #79, opened 2025-07-10
Allow email confirmation without being authenticated.

Kulv3r:


genesis_core/tests/functional/restapi/iam/test_users.py
```diff
@@ -13,13 +13,15 @@
...
+import uuid as sys_uuid
 
-from bazooka import exceptions as bazooka_exc
 import pytest
+from bazooka import exceptions as bazooka_exc
 
 from genesis_core.common import constants as common_c
 from genesis_core.tests.functional.restapi.iam import base
 from genesis_core.user_api.iam import constants as c
+from genesis_core.user_api.iam.dm.models import User
```
akremenetsky, line 24: **nit:** We always import modules and not particular objects, at least we are trying do it)

genesis_core/user_api/iam/api/controllers.py
```diff
@@ -212,11 +212,17 @@ def resend_email_confirmation(self, resource):
         resource.resend_confirmation_event(app_endpoint=app_endpoint)
         return resource
 
+    @actions.post
+    def force_confirm_email(self, resource):
```
akremenetsky, line 216: Should we register this action in routers?

akremenetsky (approved):

gmelikov (approved):

PR #83, opened 2025-07-24
Flexible login logic added, 1st/last names removed from registration

Kulv3r:
Depends on: https://github.com/infraguys/gcl_iam/pull/32


migrations/0022-nullable-user-fio-fcf9b5.py
```diff
@@ -0,0 +1,64 @@
+# Copyright 2016 Eugene Frolov <eugene@frolov.net.ru>
```
akremenetsky, line 1: Seems we need to rebase this changes since we are already have migration [# 22](https://github.com/infraguys/genesis_core/blob/master/migrations/0022-secret-certs-d340e8.py)

Kulv3r:
Wrong branch.

PR #89, opened 2025-07-31
ValidateMixin, validate password

electis:


genesis_core/user_api/iam/api/controllers.py
```diff
@@ -44,6 +47,33 @@ def enforce(self, rule, do_raise=False, exc=None):
...
 
+class ValidateMixin:
+    vm_min_length = 8
+    vm_not_contain: str = string.whitespace
+    vm_must_contain: list[str] = None  # [digits, ascii_uppercase, punctuation]
+    vm_regex: str = None
+
+    def validate(self, text):
+        if text is None:
+            return
```
Kulv3r, line 58: are you sure we should skip the validation, if new password (or any other validating string) is None?

genesis_core/user_api/iam/api/controllers.py
```diff
@@ -44,6 +47,33 @@ def enforce(self, rule, do_raise=False, exc=None):
...
+        if (
+            (self.vm_min_length and len(text) < self.vm_min_length)
+            or (self.vm_not_contain and set(self.vm_not_contain) & set(text))
+            or (
+                self.vm_must_contain
+                and not all(
+                    [set(part) & set(text) for part in self.vm_must_contain]
+                )
+            )
+            or (self.vm_regex and re.match(self.vm_regex, text))
```
Kulv3r, line 68: The regex check might be incorrect. `re.match()` returns a match object (truthy) on success and `None` (falsy) on failure, so the condition should be `not re.match()` when you want to raise an error on mismatch. Also, might be worth adding an explicit comment near `vm_regex: str = None` saying smth like "string MUST match this regex".

genesis_core/user_api/iam/api/controllers.py
```diff
@@ -44,6 +47,33 @@ def enforce(self, rule, do_raise=False, exc=None):
...
+        ):
+            raise ValidationErrorException
+
+    def validate_secret(self, kwargs: dict):
+        if "secret" in kwargs:
+            self.validate(kwargs["secret"])
```
Kulv3r, line 74: This class feels like a basic one. It might be better to either have a separate `ValidateSecretMixin(ValidateMixin)` class, or just use `.validate(kwargs['secret'])` directly in the code, which seems more reasonable here.

genesis_core/user_api/iam/api/controllers.py
```diff
@@ -44,6 +47,33 @@ def enforce(self, rule, do_raise=False, exc=None):
...
+            or (self.vm_not_contain and set(self.vm_not_contain) & set(text))
+            or (
+                self.vm_must_contain
+                and not all(
+                    [set(part) & set(text) for part in self.vm_must_contain]
+                )
+            )
+            or (self.vm_regex and re.match(self.vm_regex, text))
+        ):
+            raise ValidationErrorException
```
Kulv3r, line 70: Raising a bare ValidationErrorException (without message) isn't very helpful for debugging. Lets include what validation failed. Like so:
```
        if len(text) < self.min_length:
            raise ValidationError(f"Text must be at least {self.min_length} characters long")

        if self.not_contain and set(self.not_contain) & set(text):
            raise ValidationError("Text contains invalid characters")

        for required_chars in self.must_contain:
            if not set(required_chars) & set(text):
                raise ValidationError(f"Text must contain at least one of: {required_chars}")

        if self.regex_pattern and not re.match(self.regex_pattern, text):
            raise ValidationError(f"Text must match pattern: {self.regex_pattern}")
```

genesis_core/user_api/iam/api/controllers.py
```diff
@@ -44,6 +47,33 @@ def enforce(self, rule, do_raise=False, exc=None):
         return iam.enforcer.enforce(rule, do_raise, exc)
 
 
+class ValidateMixin:
+    vm_min_length = 8
+    vm_not_contain: str = string.whitespace
+    vm_must_contain: list[str] = None  # [digits, ascii_uppercase, punctuation]
```
Kulv3r, line 53: If `vm_must_contain` is a list, maybe `vm_not_contain` should also be a list? For consistency?
Also, these `vm_` prefixes are not necessary.

genesis_core/user_api/iam/api/controllers.py
electis: fixed and added error text

genesis_core/user_api/iam/api/controllers.py
electis: fixed

genesis_core/user_api/iam/api/controllers.py
electis: fixed via ValidateSecretMixin

genesis_core/user_api/iam/api/controllers.py
```diff
@@ -44,6 +47,47 @@ def enforce(self, rule, do_raise=False, exc=None):
...
+            exc = ValidationErrorException()
+            exc.message = error
+            raise exc
+
+
+class ValidateSecretMixin(ValidateMixin):
+    min_length = 8
+    not_contain = string.whitespace
+    must_contain = None
+    regex = None
```
Kulv3r, line 84: What's the point of copy-pasting these class attrs?

genesis_core/user_api/iam/api/controllers.py
```diff
@@ -44,6 +47,47 @@ def enforce(self, rule, do_raise=False, exc=None):
...
+            for required in self.must_contain:
+                if not set(required) & set(value):
+                    error = f"Value must contain one of {required}"
+                    break
+        elif self.regex and not re.match(self.regex, value):
+            error = f"Value must match regex {self.regex}"
+        if error:
+            exc = ValidationErrorException()
+            exc.message = error
+            raise exc
```
Kulv3r, line 77: why not:
```
if error:
    raise ValidationErrorException(error)
```

genesis_core/user_api/iam/api/controllers.py
electis: for extend future usability, this attrs for validate secret, main class attrs for validate something else..

genesis_core/user_api/iam/api/controllers.py
```diff
@@ -44,6 +47,42 @@ def enforce(self, rule, do_raise=False, exc=None):
...
+    regex: str = None
+
+    def validate(self, value):
+        error = None
+        if value is None:
+            error = "Value is required"
+        elif self.min_length and len(value) < self.min_length:
+            error = f"Value must be at least {self.min_length} characters long"
+        elif self.not_contain:
+            for not_contain in self.not_contain:
```
gmelikov, line 63: I propose to at least `set(value)` only once, because there may be many items in not_contain.

It would be really efficient if we just generate a regex from not_contain/must_contain, BUT it may be expensive to implement, so just a note for future reference.

genesis_core/user_api/iam/api/controllers.py
gmelikov: Maybe just create new exception?

Example: 
```
class ValidationPropertyIncompatibleError(ValidationErrorException):

    message = "%(val)s is not compatible with model %(model)s"

```

So you can push kwarg inside as message parameter (see examples in restalchemy)

genesis_core/user_api/iam/api/controllers.py
```diff
@@ -44,6 +47,42 @@ def enforce(self, rule, do_raise=False, exc=None):
         return iam.enforcer.enforce(rule, do_raise, exc)
 
 
+class ValidateMixin:
+    min_length = 8
```
gmelikov, line 51: It looks like private properties, let's name them as `__validate_min_length__`

And we should add `validate_` because it's a mixin, target class will have many mixins and etc.

electis: fixed

genesis_core/user_api/iam/api/controllers.py
electis: ok. at first I started doing everything through regexp, but it somehow turned out ugly.. I can redo it when I have two hands)

genesis_core/user_api/iam/api/controllers.py
```diff
@@ -44,6 +47,42 @@ def enforce(self, rule, do_raise=False, exc=None):
...
+        elif self.min_length and len(value) < self.min_length:
+            error = f"Value must be at least {self.min_length} characters long"
+        elif self.not_contain:
+            for not_contain in self.not_contain:
+                if set(not_contain) & set(value):
+                    error = f"Value must not contain {self.not_contain}"
+                    break
+        elif self.must_contain:
+            for required in self.must_contain:
+                if not set(required) & set(value):
```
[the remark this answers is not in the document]
electis: ok

genesis_core/user_api/iam/api/controllers.py
```diff
@@ -26,6 +28,7 @@
 from restalchemy.api import controllers
 from restalchemy.api import resources
 from restalchemy.common import contexts
+from restalchemy.common.exceptions import RestAlchemyException
```
akremenetsky, line 31: **nit:** We always import modules and not particular objects, at least we are trying do it)

genesis_core/user_api/iam/api/controllers.py
```diff
@@ -44,6 +47,52 @@ def enforce(self, rule, do_raise=False, exc=None):
...
+            self.__validate_regex__, value
+        ):
+            error = f"Value must match regex {self.__validate_regex__}"
+        if error:
+            raise ValidationException(description=error)
+
+
+class ValidateSecretMixin(ValidateMixin):
+    def validate_secret(self, kwargs: dict):
+        if "secret" in kwargs:
```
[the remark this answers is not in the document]
electis: in update it is not required

gmelikov (approved):

PR #98, opened 2025-08-26
Captcha (Altcha) support, check on user create

electis:


Kulv3r (approved):

genesis_core/user_api/iam/api/controllers.py
```diff
@@ -33,12 +37,49 @@
...
+        "captcha_max_number",
+        default=100000,
+        help="Captcha max number",
+    ),
+]
+
+DOMAIN = "iam_api"
+CONF = cfg.CONF
+
+CONF.register_cli_opts(
```
gmelikov, line 78: Please move configs and their declarations in appropriate place https://github.com/infraguys/genesis_core/blob/master/genesis_core/cmd/user_api.py , controller module may use config but it should not register it.

genesis_core/user_api/iam/api/controllers.py
```diff
@@ -47,6 +88,27 @@ def enforce(self, rule, do_raise=False, exc=None):
         return iam.enforcer.enforce(rule, do_raise, exc)
 
 
+class AltchaMixin:
```
gmelikov, line 91: I propose to use straightforward name instead of implementation-specific name
```suggestion
class CaptchaMixin:
```

genesis_core/user_api/iam/api/controllers.py
```diff
@@ -47,6 +88,27 @@ def enforce(self, rule, do_raise=False, exc=None):
...
+
+    def check_captcha(self, payload: str = None):
```
gmelikov, line 93: We already have `self` here, why don't get `self.request.headers.get("payload")` inside and make this method usage easier, i.e. without always copy-pasted arguments?

genesis_core/user_api/api/app.py
```diff
@@ -60,6 +60,9 @@
         "/actions/confirm_email/invoke",
         methods=[ra_c.POST],
     ),
+    iam_mw.EndpointComparator(
+        f"/v1/iam/clients/({ra_types.UUID_RE_TEMPLATE})/actions/get_captcha"
```
Kulv3r, line 64: All action pathes end with "/invoke", lets follow this convention

PR #106, opened 2025-09-12
NodeSet implementation

akremenetsky:
# Introduced NodeSet

[Documentation](https://github.com/infraguys/genesis_core/wiki/NodeSets)

A NodeSet is a group of ordinary nodes launched from the same image.
Other characteristics of nodes are also the same. Numbers of cores,
memory amount, root disk size and so on. The number of nodes in the set
is controlled by `replicas` field. Different sets types can interpretate
this field differently but the default set use it as a number of nodes
in the set.

This patch added only default set type.

Example how to create a set:
```bash
curl --location 'http://10.20.0.2:11010/v1/compute/sets/' \
--header 'Content-Type: application/json' \
--header 'Authorization: Bearer <token>' \
--data '{
    "uuid": "00000000-0000-0000-0000-000000000001",
    "name": "set",
    "project_id": "00000000-0000-0000-0000-000000000000",
    "root_disk_size": 10,
    "cores": 1,
    "ram": 1024,
    "replicas": 3,
    "image": "http://10.20.0.1:8080/genesis-base.raw"
}'
```
This request creates a set with count of replicas=3. That means under the
hood three nodes will be created.

The following component of set were added:
- User API (under /v1/computes/sets)
- Set Builder based on the Universal Builder
- A new driver for Universal Agent to work with nodes
- Bunch of tests


migrations/0030-compute-node-sets-b1869c.py
```diff
@@ -0,0 +1,76 @@
...
+                "description" varchar(255) NOT NULL,
+                "cores" integer NOT NULL,
+                "ram" integer NOT NULL,
+                "root_disk_size" integer NOT NULL,
+                "replicas" integer NOT NULL,
+                "image" varchar(255) NOT NULL,
+                "node_type" VARCHAR(2) NOT NULL CHECK (node_type IN ('VM', 'HW')),
+                "set_type" VARCHAR(32) NOT NULL CHECK (set_type IN ('SET')),
+                "status" VARCHAR(32) NOT NULL CHECK (status IN ('NEW', 'SCHEDULED', 'IN_PROGRESS', 'STARTED', 'ACTIVE', 'ERROR')),
+                "nodes" varchar(1024) NOT NULL,
```
gmelikov, line 51: I'm thinking how we will differ nodes, what do you think to add `node_set` relationship to nodes themselves? Or you think it's not explicit relationship and just optional?

gmelikov (approved):

phantomii (approved):

PR #192, opened 2025-11-26
Manifest: introduce upgrade action

gmelikov:


migrations/0037-manifest-indexes-815825.py
```diff
@@ -0,0 +1,95 @@
...
+            ALTER TABLE em_imports
+            ADD CONSTRAINT em_imports_unique_name UNIQUE (element, name);
+            """,
+            """
+            CREATE INDEX IF NOT EXISTS em_exports_element_idx
+                ON em_exports (element);
+            """,
+            """
+            ALTER TABLE em_exports
+            ADD CONSTRAINT em_exports_unique_name UNIQUE (element, name);
```
gmelikov: I've added such constraints just for ease of development. Ready to discuss it.

migrations/0037-manifest-indexes-815825.py
```diff
@@ -0,0 +1,95 @@
...
+            """,
+            """
+            CREATE INDEX IF NOT EXISTS em_resources_element_idx
+                ON em_resources (element);
+            """,
+            """
+            ALTER TABLE em_resources
+            ADD CONSTRAINT em_resources_unique_name UNIQUE (element, name);
```
gmelikov: I've added such constraints just for ease of development. Ready to discuss it.

genesis_core/elements/builders/service.py
```diff
@@ -0,0 +1,104 @@
...
+from gcl_sdk.paas.services import builder
+from gcl_sdk.agents.universal.dm import models as ua_models
+
+from genesis_core.elements.dm import res_models
+from genesis_core.elements import constants as cc
+
+LOG = logging.getLogger(__name__)
+
+
+AGENT_UUID5_NAME = "service"
```
akremenetsky: Unused?

genesis_core/elements/dm/models.py
```diff
@@ -164,88 +173,180 @@ class Manifest(
...
+    def upgrade(self):
+        element = Element.objects.get_one_or_none(
+            filters={"name": ra_filters.EQ(self.name)}
+        )
+        if not element:
+            raise ValueError(
+                f"Element '{self.name}' does not exist, please install it first."
+            )
+
+        if version.parse(element.version) >= version.parse(self.version):
```
akremenetsky: Seems it's a valid case when we would like to downgrade element version. For instance, we upgraded element, found out something  went wrong and upgraded to previous version.

genesis_core/elements/dm/models.py
```diff
@@ -164,88 +173,180 @@ class Manifest(
...
 
+        element_engine.load_from_database()
+
+        element.version = self.version
+        element.api_version = self.api_version
+        element.description = self.description
+        element.save()
+        return self.apply_element(element)
+
+    def apply_element(self, element):
```
akremenetsky: Do we handle the `requirements` section and common info such as description?

genesis_core/elements/dm/models.py
akremenetsky: Correct, I believe we need to remove this check.

genesis_core/elements/dm/models.py
gmelikov: Looks like requirements are not implemented at all now https://github.com/infraguys/genesis_core/blob/master/genesis_core/elements/dm/models.py#L345

Previous `install` didn't handle them too.

phantomii (approved):

PR #231, opened 2026-01-29
LB: introduce core_agent type

gmelikov:


genesis/images/install.sh
```diff
@@ -99,7 +99,36 @@ sudo cp "$GC_ART_DIR/undionly.kpxe" /srv/tftp/bios/undionly.kpxe
...
+# Prepare nginx for LB
+sudo mkdir -p /etc/nginx/ssl
+sudo chown www-data:www-data /etc/nginx/ssl
+sudo mkdir -p /etc/nginx/genesis/
+
+# Cert to restrict default_server
+sudo openssl req -x509 -nodes -days 3650 -newkey rsa:2048 -subj "/C=PE/ST=Genesis/L=Genesis/O=Genesis core dummy cert. /OU=IT Department/CN=genesis.core" -keyout /etc/nginx/ssl/nginx.key -out /etc/nginx/ssl/nginx.crt
+
+# Block any connections not explicitly set
+cat <<EOF | sudo tee /etc/nginx/sites-enabled/default
```
akremenetsky: We already have nginx configurations [here](https://github.com/infraguys/genesis_core/tree/master/etc/nginx/sites-available)
Would it be better to put the `default` config there?

PR #293, opened 2026-04-08
Introduce realm spec, telemetry service, dns_sync service, core_lb in manifests

gmelikov:


.vscode/launch.json
```diff
@@ -9,7 +9,7 @@
             "type": "debugpy",
             "request": "launch",
             "program": ".tox/develop/bin/gc-user-api",
-            "args": ["--config-file", "etc/genesis_core/genesis_core.conf", "--config-dir", "etc/genesis_core/genesis_core.d/"],
+            "args": ["--config-file", "etc/genesis_core/genesis_core.conf.local", "--config-dir", "etc/genesis_core/genesis_core.d/"],
```
akremenetsky, line 12: Would we like to add `*.local` files to gitignore?

gmelikov: Just thought about that, let's add it into gitignore

gmelikov: oh, wait, idea was to have sane default for local config. Maybe commit .example and ignore .local?

genesis/manifests/ecosystem_instance.yaml.j2
```diff
@@ -0,0 +1,65 @@
...
+    disable_telemetry:
+      uuid: "f6e5d4c3-b2a1-4f9e-8d7c-6b5a4f3e2d1c"
+      name: "disable_telemetry"
+      project_id: "12345678-c625-4fee-81d5-f691897b8142"
+      setter:
+        kind: selector
+        selector_strategy: latest
+
+    # Stand UUID identifier
+    stand_uuid:
```
akremenetsky: `realm_uuid`?

genesis/manifests/ecosystem_instance.yaml.j2
```diff
@@ -0,0 +1,65 @@
...
+      uuid: "9f8e7d6c-5b4a-4c3d-2e1f-0a9b8c7d6e5f"
+      name: "stand_uuid"
+      project_id: "12345678-c625-4fee-81d5-f691897b8142"
+      setter:
+        kind: selector
+        selector_strategy: latest
+
+    # Stand secret for authentication
+    stand_secret:
```
akremenetsky: `realm_secret`?

genesis_core/agent/universal/drivers/secret/backend/db.py
```diff
@@ -89,9 +94,21 @@ def create(self, resource: models.Resource) -> tp.Dict[str, tp.Any]:
...
-        self.delete(resource)
-        return self.create(resource)
+        target = secret_dm.Password.from_ua_resource(resource)
+        actual = driver_dm.Password.objects.get_one(
+            filters={
+                "uuid": dm_filters.EQ(resource.uuid),
+            }
+        )
+
+        if target.default_length != actual.meta.get("default_length", 32):
```
akremenetsky: Also we need to consider `MANUAL` password case. In this case we need to update `value` for `actual` if it changes from the `target` resource.

genesis_core/cmd/bootstrap.py
```diff
@@ -264,7 +261,52 @@ def _install_core_manifest(spec: dict[str, tp.Any], core_element_name: str = "co
...
+        return
+
+    _ensure_gctl_config(spec)
+
+    os.system(
+        f"genesis --config {GCTL_CFG_DIR}/genesisctl.yaml elements install {manifest_path}"
+    )
+
+
+def _install_ecosystem_instance_manifest(
```
akremenetsky: Would we like to have a single function to install manifests?

```python
def _install_manifest(
    spec: dict[str, tp.Any],
    element_name: str,
    element_path: str,
) -> None:

...

_install_manifest(spec, "core", ...)
_install_manifest(spec, "ecosystem", ...)
```

genesis_core/cmd/bootstrap.py
```diff
@@ -453,7 +495,229 @@ def _set_core_ip_var() -> bool:
...
+        ecosystem_endpoint_var = vs_models.Variable.objects.get_one_or_none(
+            filters={"uuid": dm_filters.EQ(c.VAR_ECOSYSTEM_ENDPOINT_UUID)}
+        )
+        if not ecosystem_endpoint_var:
+            return False
+
+        # Get ecosystem endpoint from spec or use default
+        endpoint_value = spec.get("ecosystem_endpoint", "")
+        if not endpoint_value:
+            LOG.warning("No ecosystem endpoint found in spec, using empty value")
```
akremenetsky: Is it ok we set an empty string for endpoint? How the system will handle it?

genesis_core/cmd/gservice.py
```diff
@@ -54,7 +54,7 @@ def main():
 
     engines.engine_factory.configure_postgresql_factory(CONF)
 
-    service = GeneralService(iter_min_period=3)
+    service = GeneralService(iter_min_period=1)
```
akremenetsky: Is it for debug purpose?

.vscode/launch.json
akremenetsky: Seems `.local` is different and it depend on various factors starting from a host operation system. Not sure we can provide an universal local config.

genesis_core/agent/universal/drivers/secret/backend/db.py
gmelikov: While I understand the request, now 'value' is RO https://github.com/infraguys/genesis_core/blob/master/genesis_core/user_api/secret/api/controllers.py#L45 , so we can't change it for manual case. Maybe we should support it.

genesis_core/dns_sync/service.py
```diff
@@ -0,0 +1,490 @@
...
+            "Content-Type": "application/json",
+        }
+
+    # ------------------------------------------------------------------
+    # Ecosystem HTTP helpers
+    # ------------------------------------------------------------------
+
+    def _eco_get_realm(self, endpoint, stand_uuid, auth):
+        """GET /api/ecosystem/v1/realms/{stand_uuid} -> realm dict."""
+        url = f"{endpoint}/api/ecosystem/v1/realms/{stand_uuid}"
```
akremenetsky: Would we like to include `/api/ecosystem/v1/realms/` into the endpoint variable? One day we may use a separate LB for ecosystem element.

.vscode/launch.json
gmelikov: Added .local.example

genesis/manifests/ecosystem_instance.yaml.j2
gmelikov: Renamed all to stand->realm

genesis_core/cmd/bootstrap.py
gmelikov: Good idea! Done

genesis_core/cmd/gservice.py
gmelikov: No, I've made GeneralService to honor `iter_min_period`s of each service. And if we have at least one service which we want to run each second - we have to run GeneralService each second.

genesis_core/dns_sync/service.py
gmelikov: No, we use endpoint to get to any elements in those installation (for ex. dns_sync).

akremenetsky (approved):

akremenetsky (changes requested):
_

genesis_core/cmd/bootstrap.py
gmelikov: Removed this warn

slashburygin (approved):

genesis_core/dns_sync/service.py
```diff
@@ -0,0 +1,490 @@
...
+
+        with contexts.Context().session_manager():
+            creds = self._get_ecosystem_credentials()
+            if creds is None:
+                LOG.debug("DNS sync variables are not configured, skipping")
+                return
+
+            endpoint, realm_uuid, realm_secret, access_token = creds
+
+        if self._pending_future is not None:
```
akremenetsky: Would we like to swap these two snippets in this way?
```python
        if self._pending_future is not None:
            LOG.debug("Previous DNS sync request still pending, skipping")
            return

        with contexts.Context().session_manager():
            creds = self._get_ecosystem_credentials()
            if creds is None:
                LOG.debug("DNS sync variables are not configured, skipping")
                return

            endpoint, realm_uuid, realm_secret, access_token = creds
```

In this case we don't need to fetch credentials if there is a pending future.

genesis_core/gservice/service.py
akremenetsky: Let's add `iter_min_period=iter_min_period` parameter.

akremenetsky, line 172: Let's add `iter_min_period=iter_min_period` parameter.

akremenetsky, line 179: Let's add `iter_min_period=iter_min_period` parameter.

genesis_core/gservice/service.py
```diff
@@ -179,12 +181,19 @@ def __init__(self, iter_min_period=1, iter_pause=0.1):
         cfg_service = config_service.ConfigServiceBuilder()
```
akremenetsky: Let's add `iter_min_period=iter_min_period` parameter.

genesis_core/gservice/service.py
```diff
@@ -179,12 +181,19 @@ def __init__(self, iter_min_period=1, iter_pause=0.1):
...
         secret_svc = secret_service.SecretServiceBuilder()
```
akremenetsky: Let's add `iter_min_period=iter_min_period` parameter.

genesis_core/telemetry/service.py
```diff
@@ -0,0 +1,297 @@
...
+            realm_secret = self._get_variable_value(c.VAR_REALM_SECRET_UUID)
+
+            if not all([ecosystem_endpoint, realm_uuid, realm_secret]):
+                LOG.debug("Telemetry variables are not configured, skipping")
+                return
+
+            # Collect telemetry data synchronously, send asynchronously
+            data = self._collect_telemetry()
+
+        if self._pending_future is not None:
```
akremenetsky: Let's move this snippet up. It allows skip working with DB if a `future` is still working.

genesis_core/agent/universal/drivers/secret/backend/db.py
akremenetsky: We can change the password in manifests. Actully the logic is simple. If the password is `MANUAL` and the `value` in the target resource doesn't equal to the `value` in the actual resource, we should update it.

genesis_core/dns_sync/service.py
gmelikov: Good catch!

genesis_core/agent/universal/drivers/secret/backend/db.py
gmelikov: Done

akremenetsky (approved):

PR #296, opened 2026-04-14
add all resources route, update docs, add cron for creating elements inventory

slashburygin:


gmelikov:
toute

genesis_core/user_api/em/api/controllers.py
```diff
@@ -152,6 +152,16 @@ def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
 
 
+class ResourceAllController(controllers.BaseResourceController):
+    """Controller for /v1/resources/ endpoint"""
+
+    __resource__ = resources.ResourceMap.model_type_to_resource[models.Resource]
```
akremenetsky, line 159: Why we would like to use so not typical definition for `__resource__`?
Why not:
```python
    __resource__ = resources.ResourceByRAModel(
        model_class=models.Resource,
        process_filters=True,
        convert_underscore=False,
    )
```

slashburygin: because in this case there is an error
image

akremenetsky: Let's add a comment. It's not obvious.

slashburygin: done

akremenetsky (approved):

gmelikov (approved):

PR #304, opened 2026-04-20
remove required if default in Service, update spec and add openapi MANUL test

slashburygin:


docs/openapi/openapi_boot.yaml
```diff
@@ -1,7 +1,7 @@
 openapi: 3.0.3
 info:
   title: Genesis Core v1 Boot API
-  version: 0.1.dev205+g5e9177075.d20260327
+  version: 0.0.3.dev0+g5365cb965.d20260417
```
akremenetsky: Seems it will be better to update this line only for release versions.

docs/openapi/openapi_status.yaml
```diff
@@ -1,7 +1,7 @@
 openapi: 3.0.3
 info:
   title: Genesis Core v1 Status API
-  version: 0.1.dev205+g5e9177075.d20260327
+  version: 0.0.3.dev0+g5365cb965.d20260417
```
akremenetsky: Let's update only for release versions.

akremenetsky (approved):

docs/openapi/openapi_boot.yaml
slashburygin: ok, changed in all places to latest version at this moment and in the test

gmelikov (approved):

PR #311, opened 2026-04-29
add v1/imports and v1/exports

slashburygin:


genesis_core/user_api/em/api/controllers.py
```diff
@@ -174,7 +173,16 @@ class ElementExportController(
     )
 
     def __init__(self, *args, **kwargs):
-        models.element_engine.load_from_database()
+        super().__init__(*args, **kwargs)
+
+
+class ExportAllController(controllers.BaseResourceController):
```
gmelikov: So, public controller without permissions enforcement? Same for import.

slashburygin: oh, replace inheritance to PolicyBasedController, thanks

akremenetsky (approved):

gmelikov (approved):

PR #329, opened 2026-05-08
fixes for cli and editing manifest

slashburygin:


exordos_core/elements/dm/models.py
```diff
@@ -144,25 +144,21 @@ class Manifest(
     )
     requirements = properties.property(
         ra_types.Dict(),
-        read_only=True,
```
akremenetsky, line 147: What is motivation of these changes?

.gitignore
```diff
@@ -20,6 +20,7 @@ __pycache__/
 .Python
 build/
 output/
+_tmp*/
```
gmelikov, line 23: unusual path, what's here?

slashburygin: this is a temporary path, that exists only on building element's process

exordos_core/elements/dm/models.py
slashburygin: update all necessary parts of manifest in edit cmd by vim)

akremenetsky (approved):

gmelikov (approved):

PR #334, opened 2026-05-13
func_tests

slashburygin:


.github/workflows/build.yml
```diff
@@ -47,4 +51,52 @@ jobs:
...
+        run: sudo exordos compute hypervisors init
+      - name: Download build artifacts
+        uses: actions/download-artifact@v8
+        with:
+          name: build_artifacts
+          path: artifacts/
+      - name: bootstrap exordos core
+        run: |
+          VERSION=$(cat artifacts/version.txt)
+          # VERSION=0.0.9
```
akremenetsky: Let's remove unused comment

.github/workflows/build.yml
```diff
@@ -47,4 +51,52 @@ jobs:
...
+      - name: check exordos core
+        run: |
+          set -eux
+          
+          for i in {1..10}; do
+            exordos e e list
+            
+            sleep 2
+          done
+          exordos e e show core
```
akremenetsky, line 92: So we don't check status, just wait 20 sec and print resources of the `core` element, right? Would we like to check status of the `core`?

akremenetsky (approved):

.github/workflows/build.yml
```diff
@@ -47,4 +51,50 @@ jobs:
...
+          exordos compute nodes list
+          exordos compute hypervisors list
+      - name: Setup tmate session
+        if: ${{ failure() }}
+        uses: mxschmitt/action-tmate@v3
+        timeout-minutes: 5
+        with:
+          limit-access-to-actor: true
```
gmelikov:
```suggestion
      - uses: actions/checkout@v2
      - name: Setup upterm session
        uses: owenthereal/action-upterm@v1
        if: ${{ failure() }}
        with:
          limit-access-to-actor: true
          wait-timeout-minutes: 30
```

gmelikov: I had many problems with tmate, this one looks more stable

slashburygin: thanks, I added it

gmelikov (approved):

.github/workflows/build.yml
```diff
@@ -47,4 +51,51 @@ jobs:
...
+        run: |
+          set -eux
+          
+          while [ $(exordos ee l -o json -f name=core | jq -r .[0]."Status") != "ACTIVE" ]; do
+            sleep 2
+          done
+          exordos e e show core
+          exordos compute nodes list
+          exordos compute hypervisors list
+      - uses: actions/checkout@v2
```
gmelikov:
```suggestion
```

My bad, looks like v2 is wrong and it totally don't needed here.

PR #385, opened 2026-06-10
notification: wire core via stock [events] client (drop VS coupling)

gmelikov:
Use the stock gcl_sdk event client configured via the [events] section

- user_api: build the event client only when [events] enabled, otherwise a no-op DummyEventClient, so standalone core needs no event_type_mapping file
- ec-user-api.service: load /etc/exordos_core/exordos_core.d via --config-dir so the notification element can drop in an [events] override at runtime
- drop the shipped event_type_mapping.yaml (notification now delivers it); install.sh creates the exordos_core.d drop-in dir instead
- refresh stale [events] comment hints (genesis_* -> exordos_*)


exordos_core/gservice/service.py
```diff
@@ -215,13 +225,14 @@ def __init__(self, iter_min_period=3, iter_pause=0.1):
...
-            event_sender,
             em_builder,
             dns_sync,
             # non-essential services should be last
             janitor,
             telemetry,
         ]
+        if event_sender is not None:
+            self._services.append(event_sender)
         self._next_run_times = {id(s): 0 for s in self._services}
```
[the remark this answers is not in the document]
gmelikov: Not critical.

etc/systemd/ec-user-api.service
```diff
@@ -7,7 +7,7 @@ TimeoutStopSec=5
 Restart=always
 RestartSec=5s
 KillSignal=SIGINT
-ExecStart=ec-user-api --config-file /etc/exordos_core/exordos_core.conf
+ExecStart=ec-user-api --config-file /etc/exordos_core/exordos_core.conf --config-dir /etc/exordos_core/exordos_core.d
```
slashburygin, line 10: i think, this options are mutually excluded

akremenetsky (approved):

etc/systemd/ec-user-api.service
gmelikov: I've read oslo.config docs, no, look like it uses all you give to it.

slashburygin (approved):

PR #417, opened 2026-06-16
clear old dev versions of elements. Closes #287

slashburygin:


akremenetsky:
Are we sure that we would like to keep these internal maintenance things in the main repository?

akremenetsky (approved):

gmelikov (approved):

PR #443, opened 2026-07-01
Introduce project provisioning on user registration

gmelikov:


exordos/manifests/ecosystem_realm.yaml.j2
```diff
@@ -63,3 +63,12 @@ resources:
       setter:
         kind: selector
         selector_strategy: latest
+
+    # Enable auto-provisioning of personal workspace (org + project) on email confirmation
+    registration_auto_provision:
```
akremenetsky: Is it variable of a particular realm or ecosystem? Seems if we set it to `true`, it doesn't mean the ecosystem will auto provision users.

gmelikov: for particular realm. It's about realm's iam, to be exact. Maybe not the best manifest for that?

gmelikov: It's better to move it into client, enabled by default.

phantomii (approved):

slashburygin (approved):

phantomii (changes requested):
deleted

exordos_core/user_api/iam/dm/models.py
```diff
@@ -370,6 +370,12 @@ class User(
         ra_types.Boolean(),
         default=False,
     )
+    # IamClient the user registered through; drives auto-provisioning
+    # of the personal workspace on email confirmation.
+    registration_client = properties.property(
```
cassi-volkova, line 375: `registration_client` is server-owned: both create and update discard the caller-supplied value and derive it from the actual request client. On this rebased head I reproduced the mismatch through the HTTP API: a create request containing an arbitrary client UUID returned 201, while the stored user was linked to the actual default client. The resource permissions still default this field to read-write for CREATE/UPDATE, and the checked-in OpenAPI document contains neither this field nor `registration_auto_provision`. Expected: the public schema marks the field read-only or excludes it from create/update inputs. Actual: generated clients are told they may set a value that the API silently ignores. Please fix the field permissions and regenerate `docs/openapi/openapi_user.yaml`.

cassi-volkova (changes requested):
I found one API-contract issue in the new registration-client field. The provisioning behavior passed the targeted user scenarios I ran, but the public schema and the server-owned semantics are inconsistent.

PR #460, opened 2026-07-08
Retry telemetry registration quickly instead of hourly

gmelikov:


exordos_core/user_api/iam/dm/models.py
```diff
@@ -370,6 +370,12 @@ class User(
         ra_types.Boolean(),
         default=False,
     )
+    # IamClient the user registered through; drives auto-provisioning
+    # of the personal workspace on email confirmation.
+    registration_client = properties.property(
```
akremenetsky, line 375: Why not relationship?

exordos_core/user_api/iam/api/controllers.py
```diff
@@ -190,6 +193,10 @@ def create(self, **kwargs):
         )
         self.validate_secret(kwargs)
         kwargs.pop("email_verified", None)
+        kwargs.pop("registration_client", None)
```
akremenetsky, line 196: If we would like to restrict specify `registration_client` on POST, is it possible to configure `fields_permissions` in proper way?

exordos_core/user_api/iam/api/controllers.py
```diff
@@ -211,6 +218,7 @@ def filter(self, filters, **kwargs):
     def update(self, uuid, **kwargs):
         self.validate_secret(kwargs)
         kwargs.pop("email_verified", None)
+        kwargs.pop("registration_client", None)
```
akremenetsky, line 221: The same comment as above

akremenetsky:
Duplicates https://github.com/exordos/exordos_core/pull/443 ?

exordos_core/telemetry/service.py
```diff
@@ -49,6 +59,12 @@ def __init__(self, *args, **kwargs):
         self._client = bazooka.Client(default_timeout=TELEMETRY_TIMEOUT)
         self._executor = futures.ThreadPoolExecutor(max_workers=TELEMETRY_POOL_SIZE)
         self._pending_future = None
+        # Steady-state period requested by the caller (e.g. 1 hour). Until
+        # the stand registers successfully we run at TELEMETRY_RETRY_PERIOD
+        # instead; see the comment on TELEMETRY_RETRY_PERIOD above.
+        self._steady_period = self._iter_min_period
+        self._registered = False
+        self._iter_min_period = TELEMETRY_RETRY_PERIOD
```
akremenetsky, line 74: If we cannot register even with less interval. Would we like to limited attempts?

migrations/0003-Organization-Roles-4b584d.py
```diff
@@ -94,6 +94,31 @@ def downgrade(self, session):
                     LIMIT 1
                 );
             """,
+            # Organizations without an owner member (e.g. the owner user was
+            # deleted) cannot be represented in the old schema, where owner
+            # is NOT NULL. Fall back to any member, then to the oldest user.
```
slashburygin, line 99: i think, we don't change old migrations

migrations/0064-registration-client-auto-provision-b7f2d9.py
```diff
@@ -0,0 +1,69 @@
...
+import logging
+
+from restalchemy.storage.sql import migrations
+
+LOG = logging.getLogger(__name__)
+
+
+class MigrationStep(migrations.AbstarctMigrationStep):
+    def __init__(self):
+        self._depends = ["0063-add-driver-spec-connection-uri-unique-index-a1b2c3.py"]
```
cassi-volkova: [P1] Rebase or remove the duplicated registration-provisioning migration

This dependency no longer exists on the current base branch, which has `0063-add-driver-spec-connection-uri-unique-index-eb26ec.py` followed by `0064-init-repo-tables-645142.py`. The PR also carries the registration-provisioning commit already proposed in #443, so checkout/test setup now discovers two `0064` heads.

Reproduction on commit 03d6fe9f:

1. Check out the PR against current `master`.
2. Run any functional test that initializes migrations (for example the PR CI `Tests (3.12)` job).
3. Setup aborts with `HeadMigrationNotFoundException`, listing `0064-registration-client-auto-provision-b7f2d9.py` and `0064-init-repo-tables-645142.py` as competing heads; the image build fails as well.

Please drop/rebase the duplicate provisioning commit and make any remaining migration extend the current head. Until then the telemetry retry change cannot be installed or tested through the supported workflow.

cassi-volkova (changes requested):
Tested commit 03d6fe9fdb4923e84c50f7ffc617aea5fb609b9f. The PR is not currently deployable: both the image build and functional-test setup fail before the telemetry behavior can be exercised because the migration graph has two heads.

exordos_core/telemetry/service.py
gmelikov: Good point, introduced TELEMETRY_RETRY_BACKOFF_FACTOR

exordos_core/user_api/iam/dm/models.py
gmelikov: It's not needed here, and iam have many such places where honest relationship may hurt for us.

exordos_core/user_api/iam/api/controllers.py
gmelikov: It's custom iam, other fields are the same here now

slashburygin (approved):

akremenetsky (approved):

PR #462, opened 2026-07-10
add quota

slashburygin:
## Summary by Sourcery

Introduce a quota management system for core resources and expose it via the user API.

New Features:
- Add a generic quota model and mixin to enforce per-project limits on resource counts and selected aggregate fields such as node cores and RAM.
- Expose quota limit management endpoints under the user API, including CRUD operations for quota limits and a reservations summary endpoint.

Bug Fixes:
- Prevent errors when creating duplicate SSH key host resources by handling storage conflict exceptions gracefully.

Enhancements:
- Apply quota enforcement to nodes, node sets, load balancers, and secret resources through the shared quota mixin.
- Update agent documentation to use the latest Python environment targets for unit and functional tests.

Documentation:
- Document updated tox environments for running unit and functional tests in AGENTS.md.

Tests:
- Add functional tests for quota enforcement on compute and network resources, including aggregate field limits and per-project isolation.
- Add REST API functional tests covering quota limit listing, creation, retrieval, update, and deletion.

Chores:
- Add a database migration to create the quota_limits table and supporting index.
- Introduce new user API routing for quota endpoints and wire them into the main API routes.

### DIFF
--- a/exordos_core/quota/dm/models.py
+++ b/exordos_core/quota/dm/models.py
@@ -1,0 +1,202 @@
+#    Copyright 2026 Genesis Corporation.
+#
+#    All Rights Reserved.
+#
+#    Licensed under the Apache License, Version 2.0 (the "License"); you may
+#    not use this file except in compliance with the License. You may obtain
+#    a copy of the License at
+#
+#         http://www.apache.org/licenses/LICENSE-2.0
+#
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+#    WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+#    License for the specific language governing permissions and limitations
+#    under the License.
+
+import importlib
+import logging
+import typing as tp
+import uuid as sys_uuid
+
+from restalchemy.common import exceptions as ra_e
+from restalchemy.dm import filters as dm_filters
+from restalchemy.dm import models
+from restalchemy.dm import properties
+from restalchemy.dm import types
+from restalchemy.storage.sql import orm
+
+LOG = logging.getLogger(__name__)
+
+
+class QuotaExceededError(ra_e.ValidationErrorException):
+    message = "Quota exceeded for resource '%(resource_name)s' in project %(project_id)s: %(current)s > %(limit)s"
+
+    def __init__(
+        self, resource_name: str, limit: int, current: int, project_id: sys_uuid.UUID
+    ):
+        super().__init__(
+            resource_name=resource_name,
+            limit=limit,
+            current=current,
+            project_id=project_id,
+        )
+        self.resource_name = resource_name
+        self.limit = limit
+        self.current = current
+        self.project_id = project_id
+
+
+DEFAULT_QUOTA_LIMIT = 1000
+DEFAULT_QUOTA_LIMITS: tp.Dict[str, int] = {
+    "net_lb": DEFAULT_QUOTA_LIMIT,
+    "compute_sets": DEFAULT_QUOTA_LIMIT,
+    "nodes": DEFAULT_QUOTA_LIMIT,
+    "secret_passwords": DEFAULT_QUOTA_LIMIT,
+    "secret_certificates": DEFAULT_QUOTA_LIMIT,
+    "secret_rsa_keys": DEFAULT_QUOTA_LIMIT,
+    "secret_ssh_keys": DEFAULT_QUOTA_LIMIT,
+}
+DEFAULT_QUOTA_FIELD_LIMITS: tp.Dict[str, tp.Dict[str, int]] = {
+    "nodes": {"cores": 10000},
+}
+QUOTA_RESOURCE_MODELS = {
+    "net_lb": "exordos_core.user_api.network.dm.models:LB",
+    "compute_sets": "exordos_core.compute.dm.models:NodeSet",
+    "nodes": "exordos_core.compute.dm.models:Node",
+    "secret_passwords": "exordos_core.secret.dm.models:Password",
+    "secret_certificates": "exordos_core.secret.dm.models:Certificate",
+    "secret_rsa_keys": "exordos_core.secret.dm.models:RSAKey",
+    "secret_ssh_keys": "exordos_core.secret.dm.models:SSHKey",
+}
+
+
+def get_quota_resource_model(resource_name: str) -> type:
+    try:
+        module_name, class_name = QUOTA_RESOURCE_MODELS[resource_name].split(":")
+    except KeyError:
+        raise ValueError(f"Unknown quota resource: {resource_name}")
+
+    module = importlib.import_module(module_name)
+    return getattr(module, class_name)
+
+
+class QuotaModelMixin:
+    def _quota_limits(self, session) -> tp.Collection["QuotaLimit"]:
+        limits = list(
+            QuotaLimit.objects.get_all(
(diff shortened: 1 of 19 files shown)


exordos_core/quota/dm/models.py
```diff
@@ -0,0 +1,179 @@
...
+        if field_limits:
+            field_names = {limit.field_name for limit in field_limits}
+            invalid_field_names = field_names - set(self.properties.properties)
+            if invalid_field_names:
+                raise ValueError(
+                    f"Unknown quota field(s): {', '.join(sorted(invalid_field_names))}"
+                )
+
+            fields = ", ".join(sorted(field_names))
+            result = session.execute(
```
akremenetsky: Let's do more effectively. Any field in quotas is a countable value. So we can use SQL to calculate sum and return. Something like this
```sql
SELECT SUM(cores), SUM(ram) FROM nodes WHERE project_id = %s
```

akremenetsky: Something like 

```python
fields = ", ".join(f"SUM({f})" for f in sorted(field_names))
result = session.execute(
    f"SELECT {fields} FROM {self.__tablename__} WHERE project_id = %s",
    (self.project_id,),
)
row = result.fetchone()
for quota_limit in field_limits:
    current = (row[quota_limit.field_name] or 0) + getattr(
        self, quota_limit.field_name
    )
    if current > quota_limit.limit:
        raise QuotaExceededError(
            resource_name=f"{self.__tablename__}.{quota_limit.field_name}",
            limit=quota_limit.limit,
            current=current,
            project_id=self.project_id,
        )
```

exordos_core/quota/dm/models.py
```diff
@@ -0,0 +1,179 @@
...
+            return
+
+        filters = {"project_id": dm_filters.EQ(self.project_id)}
+        field_limits = [limit for limit in limits if limit.field_name]
+        count_limits = [limit for limit in limits if not limit.field_name]
+
+        if field_limits:
+            field_names = {limit.field_name for limit in field_limits}
+            invalid_field_names = field_names - set(self.properties.properties)
+            if invalid_field_names:
```
akremenetsky: I believe we don't need to stop whole system if something goes wrong in the quotas. Let's print a warning or error and continue. For instance, we can catch this error after migration, if a field is removed from the schema but it's still in the database(QuotaLimit).

exordos_core/quota/dm/models.py
```diff
@@ -0,0 +1,179 @@
...
+                if current > quota_limit.limit:
+                    raise QuotaExceededError(
+                        resource_name=f"{self.__tablename__}.{quota_limit.field_name}",
+                        limit=quota_limit.limit,
+                        current=current,
+                        project_id=self.project_id,
+                    )
+
+        if count_limits:
+            current = self.objects.count(session=session, filters=filters) + 1
```
akremenetsky: One more request to the db) Let's do single request. Something like

```sql
SELECT COUNT(uuid) AS entity_count, SUM(cores) AS cores, SUM(ram) AS ram
FROM nodes WHERE project_id = %s
```

exordos_core/user_api/quota/api/controllers.py
```diff
@@ -0,0 +1,81 @@
...
+                name="project_id",
+                openapi_type="string",
+                param_type="project_id",
+                required=False,
+            ),
+        ],
+        responses=oa_c.build_openapi_object_response({}),
+    )
+    def filter(self, project_id: str | None = None, *args, **kwargs) -> list:
+        return models.QuotaLimit.get_project_quota_summary(project_id)
```
akremenetsky: Where `get_project_quota_summary` method is defined?

exordos_core/user_api/quota/api/controllers.py
```diff
@@ -0,0 +1,81 @@
...
+from restalchemy.api import controllers as ra_controllers
+from restalchemy.api import resources
+from restalchemy.openapi import constants as oa_c
+from restalchemy.openapi import utils
+import webob
+
+from exordos_core.quota.dm import models
+
+
+class SummaryController(ra_controllers.Controller):
```
akremenetsky: The `SummaryController` doesn't inherit from any IAM controllers.

exordos_core/user_api/quota/api/controllers.py
```diff
@@ -0,0 +1,81 @@
...
+    ra_controllers.BaseResourceControllerPaginated,
+):
+    __policy_service_name__ = "quota"
+    __policy_name__ = "limit"
+
+    __resource__ = resources.ResourceByRAModel(
+        models.QuotaLimit,
+        process_filters=True,
+        convert_underscore=False,
+    )
```
akremenetsky, line 40: Seems we need validation to create a `QuotaLimit`. For instance, we need to check fields and tables.

exordos_core/quota/dm/models.py
slashburygin: gooooood, done

exordos_core/user_api/quota/api/controllers.py
slashburygin: yeee, removed it before with pain

exordos_core/user_api/quota/api/controllers.py
slashburygin: add

exordos_core/quota/dm/models.py
slashburygin: ok ,make one sql request

exordos_core/quota/dm/models.py
slashburygin: ok

LOG.exception("Invalid quota configuration for %s", self.__tablename__)
return

exordos_core/quota/dm/models.py
```diff
@@ -0,0 +1,214 @@
...
+    "compute_sets": "exordos_core.compute.dm.models:NodeSet",
+    "nodes": "exordos_core.compute.dm.models:Node",
+    "secret_passwords": "exordos_core.secret.dm.models:Password",
+    "secret_certificates": "exordos_core.secret.dm.models:Certificate",
+    "secret_rsa_keys": "exordos_core.secret.dm.models:RSAKey",
+    "secret_ssh_keys": "exordos_core.secret.dm.models:SSHKey",
+}
+
+
+def get_quota_resource_model(resource_name: str) -> type:
```
akremenetsky, line 74: Would we like to use `lru_cache` to avoid importing models every invoke?

exordos_core/quota/dm/models.py
```diff
@@ -0,0 +1,214 @@
...
+    def validate(self):
+        super().validate()
+        resource_model = get_quota_resource_model(self.resource_name)
+        if not self.field_name:
+            return
+
+        quota_property = resource_model.properties.properties.get(self.field_name)
+        if quota_property is None:
+            raise ValueError(f"Unknown quota field: {self.field_name}")
+        if not isinstance(quota_property.get_property_type(), types.Integer):
```
akremenetsky: Float ?

exordos_core/quota/dm/models.py
slashburygin: No, creating limits is not so often to make cache

exordos_core/quota/dm/models.py
slashburygin: ok, added

akremenetsky (approved):

gmelikov (approved):

PR #473, opened 2026-07-13
Introduce Border gateway

gmelikov:


migrations/0064-init-border-ec37b4.py
```diff
@@ -0,0 +1,103 @@
...
+#    WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+#    License for the specific language governing permissions and limitations
+#    under the License.
+
+from restalchemy.storage.sql import migrations
+
+
+class MigrationStep(migrations.AbstractMigrationStep):
+    def __init__(self):
+        self._depends = ["0063-add-driver-spec-connection-uri-unique-index-a1b2c3.py"]
```
cassi-volkova: [P1] Attach the Border migrations to the current migration head

The current base branch has `0063-add-driver-spec-connection-uri-unique-index-eb26ec.py` followed by `0064-init-repo-tables-645142.py`; the `a1b2c3` dependency named here is not present. This leaves the Border chain disconnected from the repository migration chain.

Reproduction on commit 44161b85:

1. Check out the PR against current `master`.
2. Run the image build or any functional test that initializes the database.
3. Migration loading aborts with `HeadMigrationNotFoundException`, reporting `0064-init-repo-tables-645142.py` and `0068-border-type-and-ips-e5c1a9.py` as competing heads. The image is not produced and functional tests never reach the Border code.

Please rebase/renumber this chain so its first migration depends on the actual current head. This must be fixed before the feature can be deployed to an isolated test environment.

cassi-volkova (changes requested):
Tested commit 44161b856345175d40cd62a638576a8a7c041174. Unit/coverage checks pass, but the supported image build and every functional-test setup fail before the Border API can be exercised. The related gcl_sdk#163 dataplane dependency is also currently blocked by a reproduced multi-Border data-loss defect.

exordos_core/user_api/iam/api/controllers.py
```diff
@@ -183,11 +184,21 @@ def get(self, uuid, **kwargs):
...
-            do_raise=True,
-            exc=iam_e.CanNotCreateUser,
+        # The anonymous token carries no permissions by design (see
+        # gcl_iam AnonDriver), so public sign-up is impossible via
+        # permission bindings alone; it is an explicit installation
+        # choice instead ([iam] allow_self_registration).
+        ctx = contexts.get_context()
+        is_anon_signup = (
+            ctx.iam_context.token_info.token_type == "anon"
+            and cfg.CONF["iam"].allow_self_registration
```
cassi-volkova: The new anonymous-signup branch is also exercised through the functional WSGI application, which does not import the CLI entry point that registers the `[iam]` options. Reproduction on this exact head: run the anonymous registration, anonymous-without-rule, and unauthenticated user-creation functional tests. Expected: the existing permission-based case succeeds and the denied cases return 403. Actual: all three requests return HTTP 500 with `NoSuchOptError` for the missing `iam` group; the same failures are present in the current Python 3.12 CI job. Please register this option through a shared configuration path used by both the service entry point and test application (or explicitly initialize it in the functional fixture) so user creation never depends on an import side effect.

migrations/0064-init-border-ec37b4.py
cassi-volkova: Rechecked the exact head 1f25ad6. A clean PostgreSQL migration run now loads one chain through 0069; rolling back to 0064 removes the Border schema, and reapplying restores the table and final type/ipsv4 columns. The supported image Build job also completes successfully. This fixes the disconnected-head failure, so I am resolving this thread.

exordos_core/user_api/iam/api/controllers.py
cassi-volkova: This is still reproducible on exact head `1f25ad6`: the fresh Python 3.14 functional job fails the same three user-creation scenarios because the WSGI application reaches this branch without the `[iam]` option registered, resulting in HTTP 500 with `NoSuchOptError`. The migration fixes do not change this configuration path, so the defect remains.

cassi-volkova: Rechecked exact head `2312145`. The three original WSGI scenarios now pass: rule-granted anonymous registration succeeds, while anonymous registration without a matching rule and unauthenticated user creation both return the expected denial instead of 500. The full 201-test unit suite also passes. Removing the config-dependent bypass restores the existing `GrantPermissionAction` contract, so this is fixed.

exordos_core/network/border/builders/paas.py
```diff
@@ -0,0 +1,125 @@
...
+            res = ua_models.Resource.objects.get_one(
+                filters={
+                    "uuid": ra_filters.EQ(instance.uuid),
+                    "kind": ra_filters.EQ(NODE_SET_KIND),
+                }
+            )
+        except ra_storage_exc.RecordNotFound:
+            return []
+        nodeset = sdk_models.NodeSet.from_ua_resource(res)
+        return list(nodeset.nodes or {})
```
cassi-volkova, line 125: The NodeSet contains replica UUID placeholders before their VMs and universal agents exist, so returning every key here violates the readiness contract in the docstring. I reproduced this from the public API on the exact head by creating a `type: core` Border while its target node could not be scheduled immediately. Expected: the Border remains safely `IN_PROGRESS` and no `border_node` is scheduled until the VM has an address and a registered agent. Actual: the IaaS builder repeatedly raises `KeyError: 'ipv4'`, while this path tries to insert a target resource for the placeholder UUID and repeatedly hits the `ua_target_resources_agent_fkey` constraint; the Border never progresses or reports an actionable error. Please filter for provisioned nodes/registered agents before returning them (and avoid reading `ipv4` until it exists).

gmelikov: later

migrations/0064-init-border-ec37b4.py
cassi-volkova: Rechecked after the force-push to exact head `a0500882`. The rebase onto the new base reintroduced the same split migration chain: both `0065-init-border-ec37b4.py` and `0065-registration-client-auto-provision-b7f2d9.py` depend on `0064-init-repo-tables-645142.py`, leaving `0069-border-type-and-ips-e5c1a9.py` and the registration migration as competing heads. A clean HTTP functional setup now fails before the request with `HeadMigrationNotFoundException`; the exact-head Build and Python 3.14 CI jobs fail for the same reason. Please attach the Border chain after the registration migration (or otherwise restore a single migration head). I am reopening this thread because the current head cannot be built or deployed.

gmelikov: Already fixed

slashburygin (approved):

exordos_core/bootstrap/defaults.py
```diff
@@ -673,3 +674,36 @@ def set_iam_default_client_secret_var(spec: dict) -> bool:
...
+    must not fail the bootstrap.
+    """
+    endpoint = spec.get("ecosystem_endpoint", "")
+    realm_uuid = spec.get("realm_uuid", "")
+    realm_secret = spec.get("realm_secret", "")
+    if not (endpoint and realm_uuid and realm_secret):
+        LOG.info("No realm identity in spec, skipping realm registration")
+        return True
+
+    url = f"{endpoint.rstrip('/')}/api/ecosystem/v1/realms/"
```
akremenetsky: Would we like to check `no registration` flags or something like that?

exordos_core/cmd/bootstrap.py
```diff
@@ -415,7 +415,8 @@ def _migrate_installed_elements_configs() -> None:
     password,
     certificate,
     paas_lb_agent,
-    repo_proxy_installed_element
+    repo_proxy_installed_element,
+    border_agent
```
akremenetsky, line 372: It's not enought for update, you need to add something like `_migrate_installed_elements_to_repo` function that will add a new driver and capablity into the config.

exordos_core/cmd/gservice.py
```diff
@@ -34,6 +34,15 @@
         default="https://repo.exordos.com/exordos-lbaas/0.1.0/exordos-lbaas.raw.zst",
         help="URL to get image for LB dataplane VM",
     ),
+    cfg.StrOpt(
+        "border-image",
+        default=(
+            "https://repo.exordos.com/exordos-border/0.1.0/exordos-border.raw.zst"
```
akremenetsky, line 40: Is it ok?
```bash
➜  ~ curl https://repo.exordos.com/exordos-border/0.1.0/exordos-border.raw.zst                                        
<html>
<head><title>404 Not Found</title></head>
<body>
<center><h1>404 Not Found</h1></center>
<hr><center>nginx/1.24.0 (Ubuntu)</center>
</body>
</html>
➜  ~ 
```

exordos_core/network/border/builders/iaas.py
```diff
@@ -0,0 +1,110 @@
...
+    ):
+        super().__init__(instance_model)
+        self._project_id = project_id
+
+    def create_infra(
+        self, instance: models.IaasBorder
+    ) -> tp.Collection[ua_models.TargetResourceKindAwareMixin]:
+        # A pinned border (node set) never provisions its own VM,
+        # whatever the type says.
+        if instance.node or instance.type.kind != "core":
```
akremenetsky, line 60: What is `instance.node`? Is it external node, for instance external VPS or it's an ordinary platform node?

exordos_core/cmd/gservice.py
gmelikov: yep, we don't have border VM yet and won't need in this month

exordos_core/network/border/builders/iaas.py
gmelikov: ordinary core node. I made border the same as LB - you may run VM or on core agent (core agent only is used and implemented now)

exordos_core/bootstrap/defaults.py
gmelikov: Removed

akremenetsky (approved):

PR #475, opened 2026-07-13
fix(repo): filter empty optional fields from installed manifest

akremenetsky:
Empty optional fields (exports: {}, imports: {}, etc.) in the repo element manifest caused constant updates in the universal agent due to a diff between target and actual resources on every iteration.

- Add MANIFEST_OPTIONAL_FIELDS to InstalledManifest and filter empty values in from_repo_element
- Simplify _make_installed_manifest to use a loop over the shared MANIFEST_OPTIONAL_FIELDS tuple instead of individual if-blocks
- Add regression tests for empty and non-empty optional field handling

### DIFF
--- a/exordos_core/cmd/bootstrap.py
+++ b/exordos_core/cmd/bootstrap.py
@@ -347,6 +347,10 @@ def _ensure_bootstrap_repo(manifests_dir: str) -> repo_models.Repository:
 MIGRATION_REPO_NAME = "migration-dummy-repo"
 CORE_CONFIG_PATH = "/etc/exordos_core/exordos_core.conf"
 UA_CONFIG_PATH = "/etc/exordos_universal_agent/exordos_universal_agent.conf"
+CORE_CONFIG_DATA_PATH = "/var/lib/exordos/data/etc/exordos_core/exordos_core.conf"
+UA_CONFIG_DATA_PATH = (
+    "/var/lib/exordos/data/etc/exordos_universal_agent/exordos_universal_agent.conf"
+)
 
 _LAUNCHPAD_SECTION = """\
 [launchpad]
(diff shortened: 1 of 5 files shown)


exordos_core/cmd/bootstrap.py
```diff
@@ -382,6 +401,7 @@ def _migrate_installed_elements_configs() -> None:
             with open(CORE_CONFIG_PATH, "w", encoding="utf-8") as f:
                 f.write(content)
             LOG.info("Added [launchpad] section to %s", CORE_CONFIG_PATH)
+            _sync_config_to_data_path(content, CORE_CONFIG_DATA_PATH)
```
cassi-volkova, line 404: Please make a failed mirror sync retryable before the migration is marked complete. Reproduction on this exact head: make the persistent mirror temporarily unwritable and run the legacy migration, then restore write access and run it again. Expected: the mirror is reconciled before the migration is considered complete. Actual: the live config is updated, this helper only logs the write failure, and the caller continues to create the migration repository. The next bootstrap skips the whole migration because that repository exists; even invoking the config helper again skips this call because the live config already contains `[launchpad]`. I reproduced the smaller retry path directly: the live source gained the section, while the mirror remained absent both after the failed write and after a second run. Propagating the failure and/or reconciling the mirror independently of whether the live content changed would prevent the persistent config from staying stale.

cassi-volkova (dismissed):
The manifest normalization itself is consistent in the focused checks, but the newly added persistent-config synchronization has a retryability defect that can leave an upgraded installation with a stale mirror.

exordos_core/cmd/bootstrap.py
akremenetsky: We aren't not going to invest in this code much efforts. It will be deleted after migration so the current implementation is absolutely enough. If something goes wrong, just restore from backup and repeat the migration one more time.

gmelikov (approved):

exordos_core/cmd/bootstrap.py
cassi-volkova: Restoring an entire pre-migration backup can provide a recovery path, but the current code never reports that recovery is required: it catches the mirror write error, logs a warning, and still creates the migration marker. A normal retry then skips, as the reproduction showed. The PR description and migration documentation also do not state that a full backup restore is required for this failure. Since the head has not changed, the stale-mirror condition is still present, so the original review remains applicable. If this risk is deliberately accepted for the one-time migration, please document that decision and recovery procedure explicitly.

slashburygin (approved):

phantomii (approved):

PR #477, opened 2026-07-13
value_error-> validate_error

slashburygin:


cassi-volkova (changes requested):
The exception contract is changed here, but the repository's own uninstall tests still require `ValueError`. On this exact head, `tox -e develop` finishes with 189 passing tests and two failures in `test_repo_element_deps_binding.py`; the Python 3.10/3.12/3.14 and coverage jobs fail for the same reason.

Expected: the test suite and method documentation consistently assert the new `ValidateException` contract for every converted validation branch. Actual: both uninstall error scenarios raise `ValidateException` while their tests still assert `ValueError`, leaving the PR's required test matrix red. Please update the affected assertions and add/adjust coverage for the other converted install, upgrade, edit, driver-selection, and upload branches so the new public error behavior is intentional and protected.

exordos_core/repo/dm/models.py
```diff
@@ -235,7 +236,9 @@ def load_driver(self) -> "AbstractProxyRepoDriver":
                 # Just try another driver
                 pass
 
-        raise ValueError(f"Driver for spec '{self.driver_spec}' not found")
+        raise exceptions.ValidateException(
```
cassi-volkova, line 238: The public exception contract is still documented as `ValueError` here and in `Repository.upload()` and `RepoElement.edit()`, but exact-head execution confirms these paths now raise `ValidateException`, which is not a `ValueError` subclass. Please update those `Raises` sections so callers are not told to catch the wrong exception.

exordos_core/repo/dm/models.py
```diff
@@ -466,7 +471,7 @@ def dependencies(self) -> dict[str, dict[str, str]]:
 
     def install(self) -> "RepoElement":
         if self.installation_state != RepoElementInstallationState.UNINSTALLED:
-            raise ValueError("Element must be uninstalled")
```
akremenetsky, line 469: It breaks https://github.com/exordos/exordos_core/blob/master/exordos_core/cmd/bootstrap.py#L674

slashburygin: fixed, thank you very much, we need ai reviewers

PR #488, opened 2026-07-16
fix(gservice): remove unused PoolAgentDriver universal agent

VasilyStepanov:
The machine_pool_agent service wired PoolAgentDriver into a UniversalAgentService but nothing consumed its capability, and no docs or tests reference it, so drop the dead wiring from GeneralService.


VasilyStepanov:
Will be merged in https://github.com/exordos/exordos_core/pull/494

PR #494, opened 2026-07-17
feat: move VM provisioning from exordos_core into gcl_sdk

VasilyStepanov:
- Pool driver specs are kind-discriminated — MachinePool.driver_spec validates against the right schema per kind (e.g. libvirt, exordos_local_hyper) instead of being a free-form dict.
- The libvirt driver itself no longer lives here — moved to gcl_sdk; exordos_core doesn't run it or depend on libvirt-python anymore, and the now-unused PoolAgentDriver wiring was dropped from
GeneralService.
- Each hypervisor node gets its own encryption key, provisioned idempotently — MachinePool.insert() generates and stores a NodeEncryptionKey for the pool's node (check-then-create, so a node that's
both a compute Node and a hypervisor pool doesn't conflict with itself), exposed via a get_agent_private_key hypervisor action so a local agent can fetch and install the matching key.

## Summary by Sourcery

Integrate VM pool driver specifications, statuses, and boot enums from gcl_sdk, removing local pool driver implementations and wiring exordos_core to the shared SDK-based provisioning workflow.

New Features:
- Provision node-specific encryption keys for hypervisor pools during bootstrap when using the local hypervisor driver kind.

Enhancements:
- Switch compute, networking, boot, and scheduler logic to use gcl_sdk pool driver models and enums instead of exordos_core-local constants.
- Simplify hypervisor driver spec handling by dropping unused optional fields before validation to keep schemas kind-specific.
- Update gservice to use SDK-provided pool agent drivers rather than local implementations.

Build:
- Replace direct libvirt-python dependency with the gcl_sdk[libvirt] extra and align the dev image installer to install the SDK with libvirt support.

Tests:
- Adapt unit, functional, and manual tests to the new gcl_sdk-based pool driver models, statuses, and boot enums, removing tests for the deprecated local libvirt driver.


VasilyStepanov:
Run `uv lock` after https://github.com/exordos/gcl_sdk/pull/172 merge

exordos_core/user_api/compute/api/routes.py
```diff
@@ -54,11 +54,19 @@ class NodeRoute(routes.Route):
...
+
+    __controller__ = controllers.HypervisorsController
+
+
 class HypervisorRoute(routes.Route):
     """Handler for /v1/compute/hypervisors/ endpoint"""
 
     __controller__ = controllers.HypervisorsController
 
+    get_agent_private_key = routes.action(HypervisorAgentPrivateKeyActionRoute)
```
VasilyStepanov: For the bootstrap flow, I had the CLI generate the key itself upfront (generate_node_private_key_base64()), write it to the agent's disk, and pass it through the stand spec so the core's own DB-seeding
  (MachinePool.insert(agent_private_key=...)) stores the exact same key. No HTTP round-trip needed there — the CLI already knows the key because it made it.
  
  But that only works for bootstrap, where the CORE seeds the MachinePool directly into its DB during its own first boot. The other registration path — exordos compute hypervisor add (used by init --add,
  adding a hypervisor to an already-running core) — goes through the HTTP API (POST /v1/compute/hypervisors/) instead. In that flow the CLI doesn't pre-generate a key; MachinePool.insert() falls back to
  auto-generating one server-side (the agent_private_key is None branch), so the key only exists on the core's side at that point.
    
  For that agent to ever talk to status_api/orch_api with orch_secure_communication = True, the CLI needs a way to pull that server-generated key back down and write it locally before starting the agent.
  get_agent_private_key is that retrieval action — an authenticated (IAM) action on HypervisorsController, mirroring the existing NodesController.get_private_key action used for the same purpose on the
  generic Node model.
    
It's not wired into any CLI command — right now it only serves the bootstrap path, which doesn't need it. It's there so the init --add path has a real mechanism to use once that's implemented.

exordos_core/compute/dm/models.py
```diff
@@ -20,6 +20,12 @@
 
 from gcl_sdk.agents.universal.api import crypto as ua_crypto
 from gcl_sdk.agents.universal.dm import models as ua_models
+from gcl_sdk.agents.universal.drivers.pool import AbstractPoolDriverSpec  # noqa: F401
```
akremenetsky: Please use import modules instead of import attributes. It's related to other files as well.

exordos_core/gservice/service.py
```diff
@@ -100,19 +99,6 @@ def __init__(self, iter_min_period=3, iter_pause=0.1):
         volume_builder = volume_builder_svc.VolumeBuilderService(
             iter_min_period=iter_min_period
         )
-        pool_driver = ua_pool_drivers.PoolAgentDriver(
```
akremenetsky, line 106: Let's keep the agent in the core for local dev realms and for other clouds.

exordos_core/compute/dm/models.py
```diff
@@ -174,57 +110,56 @@ class MachinePool(
...
+                # key (e.g. bootstrap) must end up in sync with the DB -
+                # an existing key from unrelated prior provisioning (e.g.
+                # this node's own compute Node) would otherwise silently
+                # win, leaving the agent unable to authenticate.
+                existing_keys[0].private_key = agent_private_key
+                existing_keys[0].save(session=session)
 
-        if driver_key in self.__driver_map__:
-            return self.__driver_map__[driver_key]
+    def get_agent_private_key(self):
```
akremenetsky: Seems it's a common story and should moved to UA API

exordos_core/compute/dm/models.py
VasilyStepanov: fixed

exordos_core/gservice/service.py
VasilyStepanov: fixed

exordos_core/compute/dm/models.py
VasilyStepanov: fixed

phantomii:
LGTM

akremenetsky (approved):

gmelikov (approved):

PR #517, opened 2026-07-22
feat(ua): let external agents register and issue their node key

VasilyStepanov:
- POST /v1/ua/agents/ now accepts creation (previously the route only allowed GET/FILTER), so an agent can register with its own uuid, node, and capabilities.
  - Adds an issue_key action on that resource (/v1/ua/agents/<uuid>/actions/issue_key/invoke), gated by its own agent.ua.issue_key IAM permission. It returns the node's encryption key, generating one
  only if none exists yet — so several agents registered on the same node all get back the same key, and the operation is race-safe if two of them ask at once.
  - Seeds the new agent.ua.create / agent.ua.issue_key permissions via a migration.
  - Fixes Node.insert() to reuse an existing key for its uuid instead of unconditionally creating a new one — needed because a machine can be both a plain compute Node and a local hypervisor's node,
  and both provision a key for the same uuid; previously the second registration would crash on the unique constraint.
  - Bootstrap's existing trusted key-delivery path is untouched; this adds a separate, permissioned self-service path for agents that aren't provisioned through bootstrap.
  
  Covered by new functional tests (registration, key issuance, key sharing across agents on one node, permission enforcement) and unit tests.


VasilyStepanov:
Should be merged after https://github.com/exordos/gcl_sdk/pull/182

migrations/0069-agent-registration-key-permissions-cd5272.py
```diff
@@ -0,0 +1,83 @@
...
+#    under the License.
+
+import uuid as sys_uuid
+
+from restalchemy.storage.sql import migrations
+
+NS_UUID = sys_uuid.UUID("dfd0c604-607f-4260-981f-374f88435ea0")
+
+# UA agent permissions
+UA_AGENT_PERMISSIONS = (
```
akremenetsky: Let's do it via the core manifest. We are going to remove permission data from migrations.

VasilyStepanov: fixed

akremenetsky (approved):

phantomii (approved):

PR #523, opened 2026-07-22
Align IAM binding agent kinds with the manifest schema

### LINKED ISSUE #522: Align IAM binding agent kinds with the manifest schema
## Summary

Element manifests can declare `$core.iam.permission_bindings`, but the resulting resources cannot be scheduled because the Core agent registers a different resource kind.

## Reproduction

1. Deploy an element manifest containing a role, a permission, and a `$core.iam.permission_bindings` resource linking them.
2. Wait for element reconciliation.
3. Inspect the element resources and universal scheduler logs.

## Expected result

The permission-binding resource becomes `ACTIVE`, and the IAM permission binding exists.

## Actual result

The role and permission become `ACTIVE`, while the permission-binding resource remains `NEW`. The scheduler repeatedly reports:

```text
No agent for resource em_core_iam_permission_bindings
```

No IAM permission binding is created.

## Root cause evidence

The manifest schema and bundled example use `$core.iam.permission_bindings`, which produces the kind `em_core_iam_permission_bindings`.

However, `etc/exordos_core/core_agent.conf.j2` registers:

```ini
em_core_iam_permissionbinding = exordos_core.user_api.iam.dm.models:PermissionBinding
```

The adjacent role-binding registration has the same singular naming pattern. The canonical manifest kinds should be registered (with compatibility aliases if existing deployments rely on the old names).

Reproduced on a current development build based on Exordos Core 0.2.5. The issue is independent of any specific element payload.

cassi-volkova:
## Summary

- register the canonical `em_core_iam_permission_bindings` and `em_core_iam_role_bindings` kinds emitted by the manifest schema
- retain the legacy singular kinds as compatibility aliases
- migrate persisted core-agent configuration during bootstrap and restart only `ec-core-agent` when a change is needed
- update the bundled IAM example to use the canonical role-binding resource path

Closes #522.

## Validation

- `pytest -q -n 10 exordos_core/tests/unit` — 222 passed
- `ruff check .` — passed
- focused migration tests — 3 passed
- targeted mypy reported only existing errors outside the changed lines


cassi-volkova:
Closing this approach. The IAM binding resource kinds remain the existing singular `rolebinding` and `permissionbinding` kinds; the consuming manifest will use those canonical names directly.

PR #527, opened 2026-07-24
fix(drivers): LibvirtPoolDriver interface-type selection

VasilyStepanov:
- The transient boot-network port (Port.from_boot_network(), identified via nc.BOOT_NETWORK_PORT_UUID) always uses iface_type="network", regardless of the hypervisor's configured network_type. This avoids libvirt treating the boot network's (potentially long) logical name as a literal host bridge device name, which previously failed with Network interface name '...' is too long on bridge-type hypervisors.
- Every other port — the real port attached via attach_port(), and the real port(s) create_machine() receives when recreate_machine() rebuilds the domain post-flash — honors the hypervisor's network_type (bridge or network) as before.
- _domain2machine()/_list_interfaces() read each interface's source by its own actual type attribute rather than assuming the hypervisor's network_type applies uniformly, since the boot port and the real port can now differ in type.

### DIFF
--- a/exordos_core/compute/pool/drivers/libvirt.py
+++ b/exordos_core/compute/pool/drivers/libvirt.py
@@ -41,6 +41,25 @@ MAX_VOLUME_INDEX = 4096
 CONSOLE_LOG_DIR = "/var/log/libvirt/qemu"
 
 
+def _interface_source(
+    iface: ET.Element, source_el: tp.Optional[ET.Element]
+) -> tp.Optional[str]:
+    """Return the logical network/bridge name this interface connects to.
+
+    A domain interface created as type='network' backed by a libvirt
+    network with <forward mode='bridge'/> gets rewritten by libvirt in
+    the live/persistent XML to type='bridge', but its <source> element
+    keeps both the original 'network' attribute (the logical name the
+    orchestrator/target state actually tracks) and the 'bridge' one
+    (the real underlying device). Prefer 'network' when present so a
+    boot port riding such a network still reports its logical name
+    instead of the bridge it happens to be implemented with.
+    """
+    if source_el is None:
+        return None
+    return source_el.get("network") or source_el.get(iface.get("type"))
+
+
 class StoragePoolType(enum.Enum):
     DIR = "dir"
     ZFS = "zfs"
@@ -621,11 +640,7 @@ class LibvirtPoolDriver(base.AbstractPoolDriver):
             mac_el = iface.find("mac")
             source_el = iface.find("source")
             mac = mac_el.get("address")
-            # The boot-network port is always type='network' regardless of
-            # the hypervisor's own network_type (see create_machine) - read
-            # the source attribute this specific interface actually has
-            # ('network' or 'bridge'), not the hypervisor's configured type.
-            source = source_el.get(iface.get("type"))
+            source = _interface_source(iface, source_el)
 
             if not mac or not source:
                 raise ValueError(f"Interface {iface} has no mac or source")
@@ -704,9 +719,7 @@ class LibvirtPoolDriver(base.AbstractPoolDriver):
             mac_el = iface.find("mac")
             source_el = iface.find("source")
             mac = mac_el.get("address")
-            # See _domain2machine: read the source attribute this specific
-            # interface actually has, not the hypervisor's network_type.
-            source = source_el.get(iface.get("type"))
+            source = _interface_source(iface, source_el)
 
             if not mac or not source:
                 raise ValueError(f"Interface {iface} has no mac or source")
--- a/exordos_core/compute/pool/drivers/libvirt.py
+++ b/exordos_core/compute/pool/drivers/libvirt.py
@@ -1198,13 +1198,8 @@ class LibvirtPoolDriver(base.AbstractPoolDriver):
             raise
 
         # Build interface XML
-        # Always type='network': `port.source` is the logical libvirt
-        # network name the orchestrator tracks (see create_machine), not
-        # necessarily a literal host bridge device - a bridge-type
-        # hypervisor still needs it resolved through a local libvirt
-        # network that forwards onto the real bridge.
         interface_xml = XMLLibvirtInstance.interface_xml(
-            iface_type="network",
+            iface_type=self._spec.network_type,
             mac=port.mac,
             rom=self._spec.iface_rom_file,
             mtu=self._spec.iface_mtu,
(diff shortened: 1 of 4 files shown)


VasilyStepanov:
## Boot port must always use a network-type libvirt interface
  
  `LibvirtPoolDriver.create_machine()` builds the machine's initial (boot-network) interface using the hypervisor's own `network_type` (`bridge` or `network`) instead of a fixed `type='network'`. That's wrong: the boot network is always a plain libvirt virtual network, regardless of what the hypervisor's *main* network is configured as. On a bridge-type hypervisor, this makes libvirt treat the boot network's *name* (`exordos-core-boot-net`) as a literal host bridge device name - which is both semantically wrong and, in practice, too long for the kernel's `IFNAMSIZ` limit, so domain creation fails outright:
  
```
  Jul 24 12:14:19 hyper1 exordos-universal-agent[8488]: libvirt:  error : Network interface name 'exordos-core-boot-net' is too long: Numerical result out of range
  Jul 24 12:14:19 hyper1 exordos-universal-agent[8488]: 2026-07-24T12:14:19.631 MainProcess pid:8488 tid:136575509201408 ERROR gcl_sdk.agents.universal.services.agent:191 Error creating resource
  pool_machine with uuid (2ed9fd7f-4bb0-4dd8-8438-56a3077bbf91) and value ({'ram': 1024, 'boot': 'network', 'name': 'vm1', 'node': '2ed9fd7f-4bb0-4dd8-8438-56a3077bbf91', 'pool':
  '7518918d-65c2-40e4-9617-ae7f784ca090', 'uuid': '2ed9fd7f-4bb0-4dd8-8438-56a3077bbf91', 'cores': 1, 'image': 'https://repo.exordos.com/exordos-base/latest/exordos-base.raw.zst', 'port_info': {'mac':
  '52:54:00:55:f0:a1', 'ipv4': None, 'mask': None, 'source': 'exordos-core-boot-net'}, 'project_id': 'f1f14cf1-1639-40dc-b725-757506a202b4', 'machine_type': 'VM'})
  Traceback (most recent call last):
    File ".../gcl_sdk/agents/universal/services/agent.py", line 188, in _actualize_capability
      resource = self._create_resource(driver, r)
    File ".../gcl_sdk/agents/universal/services/agent.py", line 92, in _create_resource
      dp_resource = driver.create(resource)
    File ".../gcl_sdk/agents/universal/drivers/meta.py", line 540, in create
      meta_obj.dump_to_dp(**deps)
    File ".../gcl_sdk/agents/universal/drivers/pool.py", line 1291, in dump_to_dp
      self._create_machine(driver, dp_machine, pool_volumes, ports) 
    File ".../gcl_sdk/agents/universal/drivers/pool.py", line 1148, in _create_machine
      dp_machine, _ = driver.create_machine(dp_machine, volumes, ports)
    File ".../gcl_sdk/agents/universal/drivers/libvirt.py", line 1382, in create_machine
      self._create_domain(domain_spec)
    File ".../gcl_sdk/agents/universal/drivers/libvirt.py", line 607, in _create_domain
      virt_domain.create()                                                                                                                                                                              
    File ".../libvirt.py", line 1396, in create                                                                                                                                                         
      raise libvirtError('virDomainCreate() failed')                                                                                                                                                    
  libvirt.libvirtError: Network interface name 'exordos-core-boot-net' is too long: Numerical result out of range                                                                                       
```
                                                                                                                                                                                                        
`attach_port()`, used for the real (main-network) port attached later, is unaffected and continues to honor `network_type`.                                                                                                                                                     
                                                                                                                                                                                                        
  ### Why this PR, on its own, against `master` 

  This bug lives in `exordos_core/compute/pool/drivers/libvirt.py` today, but that module is also in the middle of being moved into `gcl_sdk` as part of a larger, separate refactor. Landing the fix here first - directly against `master`, before that move - keeps it a minimal, independently reviewable change at the place the bug actually originates, instead of bundling it into the larger relocation diff where it'd be harder to see and to bisect later. The equivalent fix already exists downstream in `gcl_sdk`'s copy of this driver; this PR brings the two back in sync at the source.

exordos_core/compute/pool/drivers/libvirt.py
```diff
@@ -1269,9 +1293,12 @@ def create_machine(
                 mac=port.mac,
                 rom=self._spec.iface_rom_file,
                 mtu=self._spec.iface_mtu,
-                # TODO(akremenetsky): This parameter should be taken from
-                # the network
-                iface_type=self._spec.network_type,
+                # Always type='network': `port.source` is a logical
```
akremenetsky: The assumption isn't correct. There are installations where `iface_type` is Linux bridge, for instance:

```xml
    <interface type="bridge">
      <mac address="52:54:01:71:28:86"/>
      <source bridge="br0"/>
      ...
    </interface>
```

I believe the root problem you need to bootstrap a new installation with Linux bridges:

```bash
exordos bootstrap --bridge br0 --boot-bridge br1 ...
```

The bridges should be created manually before running the bootstrap command.

VasilyStepanov: fixed

phantomii (approved):

phantomii:
Nice work — thank you for the careful handling of the boot and real-port cases, and for the focused regression coverage.

exordos_core/compute/pool/drivers/libvirt.py
```diff
@@ -1279,9 +1298,17 @@ def create_machine(
...
-                iface_type=self._spec.network_type,
+                # The boot-network port (see Port.from_boot_network()) is
+                # always a transient libvirt-managed network, regardless of
+                # the hypervisor's own network_type - unlike the real
+                # port(s) attached once the machine is flashed
+                # (recreate_machine()), which must honor it (e.g. a raw
+                # host bridge on bridge-type hypervisors).
+                iface_type=(
+                    "network"
+                    if port.uuid == nc.BOOT_NETWORK_PORT_UUID
```
phantomii, line 1324: _list_interfaces() assigns the zero UUID to every reconstructed port. That UUID is now the boot-port sentinel, so set_machine_cores, set_machine_ram, or rename_machine can round-trip a real bridge interface (br0) through create_machine() as type="network". Please preserve the actual interface type (or distinguish reconstructed ports from the transient boot port) and add a regression covering _list_interfaces() followed by update/recreate.

phantomii (changes requested):
Please preserve interface type across recreate and add regression coverage for the _list_interfaces() update/recreate path.

exordos_core/compute/pool/drivers/libvirt.py
VasilyStepanov: fixed

exordos_core/compute/pool/drivers/libvirt.py
```diff
@@ -1279,9 +1313,17 @@ def create_machine(
                 mac=port.mac,
                 rom=self._spec.iface_rom_file,
                 mtu=self._spec.iface_mtu,
-                # TODO(akremenetsky): This parameter should be taken from
-                # the network
-                iface_type=self._spec.network_type,
+                # The boot-network port (see Port.from_boot_network()) is
+                # always a transient libvirt-managed network, regardless of
```
akremenetsky, line 1317: It's not obvious for me why `boot-network port is always a transient libvirt-managed network`. From my point of view it's not correct. There are two "types" of network supported right now
- Libvirt networks
- Linux bridges

The first type is used only for a single node stand. If there are more than one node in the installation, the Linux bridges are used. And you need to specify "bridge" type during the bootstrap procedure.

Issue #271, opened 2026-03-24
Unreachable node on update at first 10-15 seconds after creation


akremenetsky:
If create a node and send an update call to it before loading SeedOS the node will be rebooted and become unreachable. It occurs due to [incorrect logic](https://github.com/infraguys/genesis_core/blob/master/genesis_core/compute/builders/pool.py#L258) in `PoolBuilderService`. The builder waits for `guest_machine` actual resource but the resource is created when SeedOS starts and launches agent inside itself.

The logic should be reworked to consider:
- legacy machines. They are machines without `guest_machine` target resource.
- Machines in the flashing procedure.
- Active machines.

PR #533, opened 2026-07-27
fix(compute): an update must not take a machine out of its install boot

gmelikov:
A new machine is created in network boot so Seed OS can flash it. Any update that reached it before it reported back — an element re-rendering the node it belongs to, for instance — fell through to the default hd0 and rebuilt the domain off a disk nothing had written yet: the VM lived seven seconds, came back with an empty 196 KB image and never booted.

Keep the network boot while the machine is still installing (the target asks for it and no actual guest machine exists yet); the image-changed path is unchanged.

## Summary by Sourcery

Ensure pool builder keeps machines in network boot while Seed OS installation is in progress and only switches to network boot when the guest image actually changes.

Bug Fixes:
- Prevent updates from switching installing machines off network boot before Seed OS has completed flashing the disk.

Tests:
- Add unit tests covering boot mode behavior during machine installation, for installed machines, and when the guest image changes.

### DIFF
--- a/exordos_core/compute/builders/pool.py
+++ b/exordos_core/compute/builders/pool.py
@@ -299,10 +299,21 @@ class PoolBuilderService(sdk_builder.CollectionUniversalBuilderService):
             # Don't swith boot mode for the core set as the update procedure
             # is performed by guest machine driver.
             if not self._is_core_machine(machine):
-                _, guest_actual = machine_guest_pair
+                guest_target, guest_actual = machine_guest_pair
+                # A machine that is still being installed has asked for a
+                # network boot and has not reported back yet. Any update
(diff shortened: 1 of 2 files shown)


slashburygin (approved):

akremenetsky:
Closes #271 partly. It doesn't cover the legacy case and nodes with custom images without agents.
Please don't merge until fixes for these cases. Otherwise we cannot resize such nodes.

phantomii (commented):
Could we also assert that the boot-network port is used while the machine is still installing? The current test verifies the boot mode, but not the port that makes that mode usable.

phantomii (changes requested):
Please add coverage for the boot-network port while the machine is still installing.

PR #532, opened 2026-07-26
Add in-memory IAM introspection and JWKS cache

cassi-volkova:
## Summary

- add a small Go reverse proxy that caches successful IAM token introspection responses by hashed access token and indexes them by token UUID for invalidation
- cache JWKS independently, coalesce concurrent misses, bound both caches with configurable LRU capacities, and cap introspection lifetime at the token expiration
- forward non-cacheable IAM requests and every request carrying `X-OTP` directly to Core
- expose a loopback-only idempotent invalidation endpoint; Core integration with this endpoint is intentionally deferred from the first implementation
- route IAM client traffic through the cache in the Core manifest and install/start the service during image deployment and upgrades
- build a stripped static binary with a temporary checksummed Go toolchain, then remove the toolchain and build caches from the image
- add example configuration with 15-second introspection and 1-minute JWKS defaults, plus race-enabled CI coverage

## Validation

- `go test -race ./...`
- `go vet ./...`
- `tox -e ruff-check`
- `bash -n exordos/images/install.sh exordos/images/bootstrap.sh`
- Markdown lint for `services/iam-cache/README.md`
- workflow YAML and example JSON parsing
- `git diff --check`
- full Core image build and in-place deployment in an isolated lab environment
- live health, routing, cache expiration, invalidation, pass-through, and service restart checks

## Lab benchmark

Warm-cache measurements used concurrency 32:

| Endpoint | Requests | Direct Core | Through cache | Upstream calls |
| --- | ---: | ---: | ---: | ---: |
| Introspection | 2,000 | 513 rps, p50 56.23 ms, p95 98.15 ms | 10,376 rps, p50 2.04 ms, p95 8.92 ms | 2,000 -> 1 |
| JWKS | 1,000 | 1,586 rps, p50 18.97 ms, p95 28.17 ms | 8,292 rps, p50 2.50 ms, p95 9.22 ms | 1,000 -> 1 |

Measured total Core-plus-cache CPU time decreased by approximately 98.2% for introspection and 94.0% for JWKS in this synthetic workload.

## Summary by Sourcery

Introduce a dedicated in-memory IAM reverse proxy service that caches introspection and JWKS responses and integrate it into the Core deployment and routing configuration.

New Features:
- Add the exordos-iam-cache Go service providing cached IAM introspection and JWKS endpoints with a separate internal cache invalidation API.

Enhancements:
[...]

PR #534, opened 2026-07-27
Revert "Add in-memory IAM introspection and JWKS cache"

slashburygin:
Reverts exordos/exordos_core#532

## Summary by Sourcery

Remove the in-repo IAM cache service and related wiring while switching to installation of the external IAM cache release binary and routing IAM traffic back through the core service.

Enhancements:
- Install the IAM cache using its standalone GitHub release installer instead of building it from the embedded Go service.
- Simplify load balancer configuration by routing IAM client traffic to the core HTTP backend instead of a dedicated IAM cache backend.
- Remove the bundled IAM cache configuration template and systemd unit in favor of the externally provided IAM cache packaging.

CI:
- Remove the Go test job for the now-removed in-repo IAM cache service.


exordos/images/install.sh
```diff
@@ -142,72 +141,48 @@ sudo systemctl enable nginx
...
-    ;;
-  arm64)
-    IAM_CACHE_GO_ARCH="arm64"
-    IAM_CACHE_GO_SHA256="52ce172f96e21da53b1ae9079808560d49b02ac86cecfa457217597f9bc28ab3"
-    ;;
-  *)
-    echo "Unsupported architecture for the IAM cache: $(dpkg --print-architecture)" >&2
-    exit 1
-    ;;
+    amd64|arm64)
```
phantomii, line 170: Please pin this checkout to the same release as the binary. Right now it runs whatever is on the default branch as root.

phantomii (changes requested):
Please address the comments before merging.

exordos/images/install.sh
slashburygin: ok, checkout to latest tag, same as for binary

phantomii: Using the latest release tag is fine for us. Please make the installed IAM cache service run under a dedicated unprivileged system user rather than root.

slashburygin: ok, i change user/group in iam-cache repo

ubuntu@ubuntu:~$ sudo systemctl cat exordos-iam-cache.service 
# /etc/systemd/system/exordos-iam-cache.service
[Unit]
Description=Exordos IAM Cache Service
After=network-online.target
Wants=network-online.target

[Service]
User=ubuntu
Group=ubuntu
TimeoutStopSec=10
Restart=always
RestartSec=5s
KillSignal=SIGINT
ExecStart=/usr/local/bin/exordos-iam-cache -config /etc/exordos-iam-cache/config.json

[Install]
WantedBy=multi-user.target

exordos/images/install.sh
```diff
@@ -142,72 +141,57 @@ sudo systemctl enable nginx
...
+    | cut -d '"' -f 4 \
+    | grep -E ".*linux[-_]${IAM_CACHE_ARCH}$" \
+    | head -n 1)
+if [[ -z "$IAM_CACHE_DOWNLOAD_URL" ]]; then
+    echo "No Linux ${IAM_CACHE_ARCH} IAM cache binary found in release ${IAM_CACHE_TAG}." >&2
+    exit 1
+fi
+
+IAM_CACHE_BINARY="$IAM_CACHE_DIR/exordos-iam-cache"
+curl -fsSLo "$IAM_CACHE_BINARY" "$IAM_CACHE_DOWNLOAD_URL"
```
phantomii, line 191: The release already provides checksums.txt, but this installer downloads and runs the binary without verifying it. Could we verify the checksum before executing the installer?

phantomii (changes requested):

exordos/images/install.sh
slashburygin: ok, verified

phantomii (approved):
Verified the latest revision: checksum verification is in place and the IAM cache service runs as an unprivileged dedicated user.

VasilyStepanov (approved):

PR #541, opened 2026-07-30
move part of entities with zero uuid to manifest

slashburygin:
## Summary by Sourcery

Introduce ZERO_UUID constant and migrate zero-UUID IAM entities from SQL bootstrap to manifest-based configuration, updating related tests, bootstrap logic, and services to reference the new constants and IDs.

New Features:
- Define core IAM organizations, projects, roles, permissions, and bindings in the core manifest instead of relying purely on SQL seed data.

Enhancements:
- Add ZERO_UUID and additional IAM-related UUID constants for shared use across migrations, tests, and services.
- Refactor code and tests to use ZERO_UUID and new IAM binding/role/permission UUID constants instead of hard-coded zero UUID strings.
- Extend core agent configuration with aliases for IAM role and permission binding models.

Tests:
- Update functional and unit tests to use ZERO_UUID and new IAM entity UUIDs, and to avoid hard-coded zero UUID assumptions in role/permission binding scenarios.

Chores:
- Add a new migration to reassign IAM admin organization, role, and allow-all permission from the zero UUID to dedicated UUIDs, with corresponding downgrade logic.


VasilyStepanov (approved):

exordos_core/user_api/iam/api/controllers.py
```diff
@@ -310,9 +310,7 @@ def _get_request_iam_client(self) -> models.IamClient | None:
             # Anonymous registration (e.g. street sign-up) is not bound to
             # a client token; attribute it to the default client.
             return models.IamClient.objects.get_one_or_none(
-                filters={
-                    "uuid": ra_filters.EQ(sys_uuid.UUID(common_c.DEFAULT_CLIENT_UUID))
-                }
+                filters={"uuid": ra_filters.EQ(sys_uuid.UUID(str(common_c.ZERO_UUID)))}
```
akremenetsky, line 313: I prefer to keep `DEFAULT_CLIENT_UUID` since it make sence semantically.

migrations/0001-zero-entities-5a7a0e.py
```diff
@@ -0,0 +1,391 @@
...
+
+
+def upgrade_admin_organization(session):
+    session.execute(
+        f"""
+        INSERT INTO iam_organizations (
+            uuid, status, name, description, created_at, updated_at, info
+        )
+        SELECT
+            '{NEW_ADMIN_ORG_UUID}',
```
akremenetsky, line 39: Is it real case where we need to change organizations UUID?

exordos/manifests/core.yaml.j2
```diff
@@ -7,14 +7,650 @@ api_version: "v1"
...
+      user: "00000000-0000-0000-0000-000000000000"
+      # user: "$core.iam.users.$admin:uuid"
+      role: "OWNER"
 
+  $core.iam.permissionbinding:
+    admin_allow_all:
+      uuid: "0fa34733-a6c3-4878-83a5-30c894d41b7c"
+      role: "$core.iam.roles.$admin:uuid"
+      permission: "$core.iam.permissions.$allow_all:uuid"
+    binding_newcomer_iam_user_delete:
```
phantomii, line 456: Most permission bindings in this section have no explicit UUID, while the same role, permission, and project combinations are already inserted by the squashed migration. Applying the manifest generates new UUIDs for them, and the table only enforces uniqueness on UUID, so both rows remain. I reproduced this with the locked gcl-sdk: 46 existing bindings get duplicates. This also means removing a binding from the manifest does not actually revoke the permission because the SQL-seeded row remains. Could we assign stable UUIDs here and migrate or remove the existing rows?

phantomii (commented):
I found two data consistency issues in this patch.

migrations/0001-zero-entities-5a7a0e.py
```diff
@@ -0,0 +1,391 @@
...
+        WHERE uuid = '{NEW_ADMIN_ORG_UUID}';
+        """
+    )
+    session.execute(
+        f"""
+        INSERT INTO iam_organizations (
+            uuid, status, name, description, created_at, updated_at, info
+        )
+        SELECT
+            '{ZERO_UUID}',
```
phantomii, line 227: This temporary name is copied into the organization restored at the zero UUID, but it is never changed back to admin. I reproduced an upgrade followed by a rollback on PostgreSQL, and the restored organization was left with the name __downgrading_admin_organization__. Could the downgrade restore the original name so the round trip preserves the organization?

akremenetsky: +1

exordos/manifests/core.yaml.j2
slashburygin: i think, this is not required

exordos_core/user_api/iam/api/controllers.py
slashburygin: In next PR i change here zero_uuid to direct not zero client uuid

migrations/0001-zero-entities-5a7a0e.py
slashburygin: ok, good idea , fixed

akremenetsky (approved):

PR #542, opened 2026-07-30
add tags to resource models

slashburygin:
## Summary by Sourcery

Enable tagging across supported resource models with persistent, queryable tag metadata.

New Features:
- Add optional string tags to resource models and expose them in the API specification.

Enhancements:
- Centralize table-to-model mappings for resource handling and quota lookups.
- Support indexed tag storage across the resource tables.

### DIFF
--- a/migrations/0003-add-tags-to-models-9e4b2d1f.py
+++ b/migrations/0003-add-tags-to-models-9e4b2d1f.py
@@ -1,0 +1,49 @@
+#    Copyright 2026 Genesis Corporation.
+#
+#    All Rights Reserved.
+#
+#    Licensed under the Apache License, Version 2.0 (the "License"); you may
+#    not use this file except in compliance with the License. You may obtain
+#    a copy of the License at
+#
+#         http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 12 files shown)


migrations/0003-add-tags-to-models-9e4b2d1f.py
```diff
@@ -0,0 +1,49 @@
...
+    @property
+    def migration_id(self):
+        return "9e4b2d1f-58cc-4372-a567-0e02b2c3d479"
+
+    @property
+    def is_manual(self):
+        return False
+
+    def upgrade(self, session):
+        for table in constants.TABLES_TO_MODELS.keys():
```
akremenetsky, line 35: What we are going to do when the `TABLES_TO_MODELS` dict is extend? I mean for installations where this migration is already applied,  new values don't have tags. We need a separate migration for them.

slashburygin: well, of course, new resources/tables will require new migrations.)

akremenetsky (approved):

phantomii (approved):

akremenetsky (approved):

PR #544, opened 2026-07-31
stable, latest, tags, other props for repo elements

slashburygin:
## Summary by Sourcery

Add support for filtering repository elements by stable versions and align repository element error handling with common validation exceptions.

New Features:
- Introduce a stable version helper and is_stable property on repository elements to distinguish pre-release versions.
- Add a stable query parameter to the repo elements REST API to filter responses to only stable versions when requested.

Enhancements:
- Replace generic ValueError usages in repository element and driver operations with domain-specific ValidateException for consistent API error semantics.

Tests:
- Add functional REST API tests for repository elements covering stable filtering, access control, lifecycle actions (install, uninstall, upgrade, edit), deletion constraints, and hidden fields behavior.

### DIFF
--- a/exordos_core/network/lb/builders/iaas.py
+++ b/exordos_core/network/lb/builders/iaas.py
@@ -53,7 +53,7 @@ class LBBuilder(builder.CoreInfraBuilder):
 
         node_set = models.TargetNodeSet(
             uuid=instance.uuid,
-            name=f"{self._name_prefix}-{instance.name or str(instance.uuid[:4])}",
+            name=f"{self._name_prefix}-{instance.name or str(instance.uuid)[:4]}",
             cores=instance.type.cpu,
             ram=instance.type.ram,
             replicas=instance.type.nodes_number,
@@ -77,7 +77,6 @@ class LBBuilder(builder.CoreInfraBuilder):
             return tuple()
         nodeset = None
         tgt_nodeset = None
-
         for target, actual in infra.infra_objects:
             if target.get_resource_kind() == NODE_SET_KIND:
                 nodeset = actual
--- a/exordos_core/repo/builders/repository.py
+++ b/exordos_core/repo/builders/repository.py
@@ -87,15 +87,13 @@ class RepoProxyBuilderService(
 
         # Save the elements to the database
         i = -1
-        saved_elements = []
         for i, element in enumerate(instance.iter_elements_in_inventory()):
             # In COPY mode, actualize element to download full data
             if instance.sync_mode == models.SyncMode.COPY:
                 instance.actualize_element(element)
             else:
                 element.save()
-            saved_elements.append(element)
-            self._set_latest_for_new_element(saved_elements, element)
+            self._set_latest_for_new_element(instance, element)
 
         LOG.info(
             "Saved %d elements to the database for repository %s",
@@ -136,8 +134,7 @@ class RepoProxyBuilderService(
                 instance.actualize_element(element)
             else:
                 element.save()
-            existing_elements[key] = element
-            self._set_latest_for_new_element(existing_elements.values(), element)
+            self._set_latest_for_new_element(instance, element)
             new_elements_count += 1
 
         if new_elements_count:
@@ -168,11 +165,8 @@ class RepoProxyBuilderService(
 
             was_latest_stable = element.stable and element.latest
             element.delete()
-            del existing_elements[key]
             if was_latest_stable:
-                self._set_latest_for_outdated_element(
-                    existing_elements.values(), element.name
-                )
+                self._set_latest_for_outdated_element(instance, element.name)
             deleted_elements_count += 1
 
         if deleted_elements_count:
@@ -191,27 +185,23 @@ class RepoProxyBuilderService(
             instance.update()
 
     def _set_latest_for_new_element(
-        self,
-        elements: tp.Collection[models.RepoElement],
-        element: models.RepoElement,
+        self, instance: models.Repository, element: models.RepoElement
     ) -> None:
         """Set `latest` when a new stable element is the greatest version."""
         if not element.stable:
             return
 
-        same_name = [
-            candidate
-            for candidate in elements
-            if candidate.stable and candidate.name == element.name
-        ]
+        elements = models.RepoElement.objects.get_all(
+            filters={"repository": instance.uuid, "name": element.name},
+        )
         latest_element = max(
-            same_name,
+            (candidate for candidate in elements if candidate.stable),
             key=lambda candidate: packaging_version.parse(candidate.version),
         )
         if latest_element.uuid != element.uuid:
             return
 
-        for candidate in same_name:
+        for candidate in elements:
             if candidate.uuid != element.uuid and candidate.latest:
                 candidate.latest = False
                 candidate.update()
@@ -221,14 +211,13 @@ class RepoProxyBuilderService(
             element.update()
 
     def _set_latest_for_outdated_element(
-        self, elements: tp.Collection[models.RepoElement], name: str
+        self, instance: models.Repository, name: str
     ) -> None:
         """Promote the greatest remaining stable version after deleting latest."""
-        stable_elements = [
-            element
-            for element in elements
-            if element.stable and element.name == name
-        ]
+        elements = models.RepoElement.objects.get_all(
+            filters={"repository": instance.uuid, "name": name},
+        )
+        stable_elements = [element for element in elements if element.stable]
         if not stable_elements:
             return
 
--- a/exordos_core/repo/builders/repository.py
+++ b/exordos_core/repo/builders/repository.py
@@ -21,6 +21,7 @@ import typing as tp
 from gcl_looper.services.oslo import base as oslo_base
 from gcl_sdk.agents.universal.dm import models as ua_models
 from gcl_sdk.agents.universal.services import builder as sdk_builder
+from packaging import version as packaging_version
 
 from exordos_core.repo.dm import models
 
@@ -92,6 +93,7 @@ class RepoProxyBuilderService(
                 instance.actualize_element(element)
             else:
                 element.save()
+            self._set_latest_for_new_element(instance, element)
 
         LOG.info(
             "Saved %d elements to the database for repository %s",
@@ -132,6 +134,7 @@ class RepoProxyBuilderService(
                 instance.actualize_element(element)
             else:
                 element.save()
+            self._set_latest_for_new_element(instance, element)
             new_elements_count += 1
 
         if new_elements_count:
@@ -160,7 +163,10 @@ class RepoProxyBuilderService(
                 )
                 continue
 
+            was_latest_stable = element.stable and element.latest
             element.delete()
+            if was_latest_stable:
+                self._set_latest_for_outdated_element(instance, element.name)
             deleted_elements_count += 1
 
         if deleted_elements_count:
@@ -177,6 +183,51 @@ class RepoProxyBuilderService(
                 seconds=instance.refresh_rate
             )
             instance.update()
+
+    def _set_latest_for_new_element(
+        self, instance: models.Repository, element: models.RepoElement
+    ) -> None:
+        """Set `latest` when a new stable element is the greatest version."""
+        if not element.stable:
+            return
+
+        elements = models.RepoElement.objects.get_all(
+            filters={"repository": instance.uuid, "name": element.name},
+        )
+        latest_element = max(
+            (candidate for candidate in elements if candidate.stable),
+            key=lambda candidate: packaging_version.parse(candidate.version),
+        )
+        if latest_element.uuid != element.uuid:
+            return
+
+        for candidate in elements:
+            if candidate.uuid != element.uuid and candidate.latest:
+                candidate.latest = False
+                candidate.update()
+
+        if not element.latest:
+            element.latest = True
+            element.update()
+
+    def _set_latest_for_outdated_element(
+        self, instance: models.Repository, name: str
+    ) -> None:
+        """Promote the greatest remaining stable version after deleting latest."""
+        elements = models.RepoElement.objects.get_all(
+            filters={"repository": instance.uuid, "name": name},
+        )
+        stable_elements = [element for element in elements if element.stable]
+        if not stable_elements:
+            return
+
+        latest_element = max(
+            stable_elements,
+            key=lambda element: packaging_version.parse(element.version),
+        )
+        if not latest_element.latest:
+            latest_element.latest = True
+            latest_element.update()
 
     def post_update_instance_resource(
         self,
--- a/exordos_core/repo/builders/repository.py
+++ b/exordos_core/repo/builders/repository.py
@@ -87,13 +87,15 @@ class RepoProxyBuilderService(
 
         # Save the elements to the database
         i = -1
+        saved_elements = []
         for i, element in enumerate(instance.iter_elements_in_inventory()):
             # In COPY mode, actualize element to download full data
             if instance.sync_mode == models.SyncMode.COPY:
                 instance.actualize_element(element)
             else:
                 element.save()
-            self._set_latest_for_new_element(instance, element)
+            saved_elements.append(element)
+            self._set_latest_for_new_element(saved_elements, element)
 
         LOG.info(
             "Saved %d elements to the database for repository %s",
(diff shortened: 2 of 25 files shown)


exordos_core/user_api/repo/api/controllers.py
```diff
@@ -103,6 +104,36 @@ class RepoElementController(
         hidden_fields=["installation_state"],
     )
 
+    def filter(self, filters, **kwargs):
```
akremenetsky: 1) Seems it breaks pagination. The stable/latest filtering is applied after super().filter() returns. When pagination is active (page_limit query param), super().filter() calls paginated_filter() which returns exactly _pagination_limit items.
2) Why we would like to place this custom logic into controller? It's only need for CLI, so let's place it into CLI. Otherwise we slow down API.
3) By default only stable version is printed in CLI. We can specify `--dev` flag to print all version

exordos_core/repo/dm/models.py
```diff
@@ -466,7 +483,7 @@ def dependencies(self) -> dict[str, dict[str, str]]:
 
     def install(self) -> "RepoElement":
         if self.installation_state != RepoElementInstallationState.UNINSTALLED:
-            raise ValueError("Element must be uninstalled")
+            raise common_exc.ValidateException(err="Element must be uninstalled")
```
akremenetsky, line 508: We broke this logic https://github.com/exordos/exordos_core/blob/master/exordos_core/cmd/bootstrap.py#L674

pyproject.toml
```diff
@@ -38,6 +38,7 @@ dependencies = [
     "firebase-admin>=6.0.0,<8.0.0",  # Apache-2.0
     "altcha>=0.2.0,<2.1.0",  # MIT License
     "openapi-schema-validator>=0.8.1",  # BSD-3
+    "packaging==26.0",  # Apache-2.0 or BSD
```
akremenetsky: We would like to pin to `packaging==26.0` exactly? Maybe range of versions?

slashburygin: ok, packaging>=26.0,<=26.3

exordos_core/repo/dm/models.py
slashburygin: fixed, thanks

exordos_core/user_api/repo/api/controllers.py
slashburygin: outdated

exordos_core/elements/dm/models.py
```diff
@@ -136,6 +136,38 @@ class Manifest(
         ra_types.AllowNone(ra_types.String(max_length=16, min_length=1)),
         default=None,
     )
+    summary = properties.property(
+        ra_types.String(),
+        default="",
+    )
+    rating = properties.property(
```
akremenetsky: We shouldn't place the `rating` field into manifest.

exordos_core/network/lb/builders/iaas.py
```diff
@@ -75,8 +75,6 @@ def actualize_infra(
     ) -> tp.Collection[ua_models.TargetResourceKindAwareMixin]:
         if instance.type.kind != "core":
             return tuple()
-        nodeset = None
```
akremenetsky, line 78: Can we get an unbound error here?

exordos_core/repo/builders/repository.py
```diff
@@ -170,6 +171,29 @@ def _refresh_repository(self, instance: models.Repository) -> None:
                 instance.repo_uri,
             )
 
+        elements = models.RepoElement.objects.get_all(
```
akremenetsky: We fetch at the begin of the method. Why we need to fetch all elements again?

exordos_core/network/lb/builders/iaas.py
slashburygin: this were unused variblaes

exordos_core/repo/builders/repository.py
slashburygin: do all in one cycle now

exordos_core/network/lb/builders/iaas.py
akremenetsky: If the collection `infra.infra_objects` is empty, the cycle passes zero times, so we get an unbound local error.

exordos_core/user_api/repo/api/controllers.py
```diff
@@ -104,7 +140,9 @@ class RepoElementController(
     )
 
     def get(self, uuid, **kwargs):
-        repo_element = super().get(uuid=uuid, **kwargs)
+        repo_element = controllers.BaseResourceControllerPaginated.get(
```
akremenetsky: Why we mask IAM methods?

exordos_core/user_api/repo/api/controllers.py
```diff
@@ -82,6 +86,38 @@ def upload(
...
+            }
+        )
+        for element in elements:
+            try:
+                version = parse_version(element.version)
+            except InvalidVersion:
+                continue
+
+            current = elements_by_name.get(element.name)
+            if current is None or (version, element.repository.priority) > current[:2]:
```
akremenetsky: It's not correct. Element from most priority repo should be selected even though its version is less. In other words, priority should be checked first.

exordos_core/user_api/repo/api/controllers.py
```diff
@@ -117,9 +155,26 @@ def delete(self, uuid):
...
+        elements = models.RepoElement.objects.get_all(
+            filters={
+                "name": dm_filters.EQ(resource.name),
+                "stable": dm_filters.EQ(True),
+                "uuid": dm_filters.NE(resource.uuid),
+            },
+            order_by={"version": "desc"},
+        )
+        return sorted(
+            (element for element in elements if element.uuid != resource.uuid),
```
akremenetsky: Why we need this condition `if element.uuid != resource.uuid`? Seems the filter is used for that `"uuid": dm_filters.NE(resource.uuid)`.

exordos_core/repo/builders/repository.py
akremenetsky: You don't need this cycle at all.
1) The stable elements is already set in `iter_elements_in_inventory`. So do nothing to detect new element stable or not. The field `stable` should have a correct value.
2) To detect latest element you may use two cycles above. For new elements and for outdated elements. 
2.1) For new elements. If the element stable, iterate over all elements with the same name and check if the version is latest. If it's true, set the field `latest` to True.
2.2) For outdated elements. If the element stable, iterate over all elements with the same name and find other element with nearest version. Set this element as latest.

exordos_core/repo/dm/models.py
```diff
@@ -255,6 +266,7 @@ def iter_elements_in_inventory(
                 element = RepoElement(
                     name=name,
                     version=version,
+                    stable=is_stable_version(version),
```
akremenetsky, line 269: We set `stable` but not `latest`. Seems for repos added in runtime the field `latest` isn't set before first refresh. It should be checked.

exordos_core/user_api/repo/api/controllers.py
slashburygin: because we want to see element without auth

exordos_core/user_api/repo/api/controllers.py
slashburygin: ye, it,s duplicated, removed

exordos_core/repo/dm/models.py
slashburygin: After small changes):
During repository creation, every inventory element is saved/actualized and then immediately passed to

self._set_latest_for_new_element(instance, element)

akremenetsky (approved):

migrations/0002-add-repo-element-version-flags-24310b.py
```diff
@@ -0,0 +1,63 @@
...
+        session.execute(
+            """
+            ALTER TABLE repo_elements
+            ADD COLUMN IF NOT EXISTS tags TEXT[] NOT NULL DEFAULT '{}',
+            ADD COLUMN IF NOT EXISTS published_at timestamp(6) without time zone DEFAULT now(),
+            ADD COLUMN IF NOT EXISTS stable boolean NOT NULL DEFAULT false,
+            ADD COLUMN IF NOT EXISTS latest boolean NOT NULL DEFAULT false;
+            """
+        )
+        session.execute(
```
phantomii, line 48: The migration leaves existing elements with both flags set to false. A regular refresh only calculates them for newly discovered versions, so existing repositories will not appear in the stable store until a new version is added. Could we backfill these fields during migration or refresh all existing elements once?

exordos_core/user_api/repo/api/controllers.py
```diff
@@ -82,6 +86,63 @@ def upload(
...
+                version = parse_version(element.version)
+            except InvalidVersion:
+                continue
+
+            current = elements_by_name.get(element.name)
+            candidate = (element.repository.priority, version, element)
+            if current is None or candidate[:2] > current[:2]:
+                elements_by_name[element.name] = candidate
+
+        return [candidate[2] for candidate in elements_by_name.values()]
```
phantomii: This endpoint is paginated, but the custom filter ignores pagination and loads every matching element before returning the response. As the store grows, even a small page request will read and return the whole catalogue. Could we apply pagination after selecting the per-name entries and add coverage for it?

migrations/0002-add-repo-element-version-flags-24310b.py
```diff
@@ -0,0 +1,63 @@
...
+            "CREATE INDEX IF NOT EXISTS idx_repo_elements_project_id ON repo_elements (project_id)"
```
phantomii, line 49: The store query filters every request by `stable` and `latest`, but this migration does not add an index for those fields. As the number of elements grows, this will require scanning the whole table before the controller can select the catalogue entries. Could we add an index that supports this lookup?

phantomii (changes requested):
Please address the comments above.

exordos_core/user_api/repo/api/controllers.py
slashburygin: add pagination

migrations/0002-add-repo-element-version-flags-24310b.py
slashburygin: CREATE INDEX IF NOT EXISTS idx_repo_elements_latest_stable
ON repo_elements (name) WHERE stable AND latest

Partial rather than a plain (stable, latest) composite: the store's only predicate is stable AND latest, so the index holds just the catalogue rows. Dropped in downgrade() alongside the other two.

migrations/0002-add-repo-element-version-flags-24310b.py
slashburygin: now migration set latest and stable flags, the same rule _set_latest_for_new_element applies during refresh.

phantomii (commented):
The latest-flag invariant is still broken for direct API write paths. `Repository.upload()` saves a stable element with `latest=False` without invoking the builder recomputation, so first/newer stable uploads can be absent from `latest_stable_elements`. `RepoElementController.delete()` also deletes the current latest without promoting a remaining stable version. Please centralize recomputation after upload and delete, with regression coverage for first/newer stable uploads and deletion of the latest element.

phantomii (commented):
I found several additional issues that need to be addressed:

- The public store exposes the full project-owned RepoElement model without a public eligibility check or a safe field projection. Anonymous callers can enumerate and fetch data from private repositories, and stable_versions mixes elements across projects.
- When an installed latest version disappears from remote inventory, refresh keeps it marked latest and it remains in the store catalogue. Keep the installed row for lifecycle purposes, but remove it from the catalogue and promote an available version.
- GET /v1/repo/store/ is broken: StoreProxyController combines the policy controller with a routes-list controller, then its filter path expects a resource model that the proxy does not have. Use a routes-list controller directly or provide an explicit route-list filter, with coverage for the root endpoint.
- latest_stable_elements ignores client filters and ordering. Apply supported filters before collapsing per name, or reject unsupported query parameters rather than returning an unfiltered catalogue.
- The migration assigns current migration time as published_at for all existing elements. Backfill from created_at (or inventory metadata) so old packages do not appear newly published, and add migration coverage.

exordos_core/user_api/repo/api/controllers.py
```diff
@@ -82,6 +86,86 @@ def upload(
...
+
+
+class StoreProxyController(
+    iam_controllers.PolicyBasedWithoutProjectController,
+    controllers.RoutesListController,
+):
+    __TARGET_PATH__ = "/v1/repo/store/"
+
+
+class StoreElementController(controllers.BaseResourceControllerPaginated):
```
phantomii: The store is public, but this controller exposes the full project-owned RepoElement model and only hides installation_state. Anonymous callers can enumerate and fetch elements from private repositories, including their manifest and repository references; stable_versions also mixes same-name elements across projects. Please define an explicit public-catalogue predicate and return a safe public projection, with cross-project negative coverage.

exordos_core/repo/builders/repository.py
```diff
@@ -160,7 +163,10 @@ def _refresh_repository(self, instance: models.Repository) -> None:
                 )
                 continue
```
phantomii, line 166: Installed elements that disappear from inventory are intentionally retained, but this branch skips their catalogue maintenance. If the disappearing row is latest, it remains latest and is still returned by the store even though the source no longer provides it. Preserve the installed row for lifecycle purposes, but clear its catalogue eligibility and promote an available stable version. Please add coverage for this path.

exordos_core/user_api/repo/api/controllers.py
```diff
@@ -82,6 +86,86 @@ def upload(
...
+        if self._pagination_marker:
+            elements = [
+                element
+                for element in elements
+                if element.uuid > self._pagination_marker
+            ]
+        return elements[: self._pagination_limit]
+
+
+class StoreProxyController(
```
phantomii: GET /v1/repo/store/ is allowed, but this MRO sends its collection request through the policy controller’s resource filter. The proxy has no resource model, so the request fails with an internal error. Use RoutesListController directly or provide an explicit route-list filter, and add a root-store endpoint test.

exordos_core/user_api/repo/api/controllers.py
```diff
@@ -82,6 +86,86 @@ def upload(
...
+class LatestStableElementsController(controllers.BaseResourceControllerPaginated):
+    __resource__ = resources.ResourceByRAModel(
+        models.RepoElement,
+        convert_underscore=False,
+        hidden_fields=["installation_state"],
+    )
+
+    def filter(self, filters, **kwargs):
+        elements_by_name = {}
+        elements = models.RepoElement.objects.get_all(
```
phantomii, line 123: This ignores the client-provided filters and ordering: get_all receives only the hard-coded stable/latest predicates, so a request such as name=<value> still returns the whole catalogue. Apply supported filters before the per-name collapse, or reject unsupported query parameters rather than returning an incorrect unfiltered response. Please add coverage for this contract.

migrations/0002-add-repo-element-version-flags-24310b.py
```diff
@@ -0,0 +1,115 @@
...
+        return "24310bc7-03bf-4ecc-af05-7199ae6d5460"
+
+    @property
+    def is_manual(self):
+        return False
+
+    def upgrade(self, session):
+        session.execute(
+            """
+            ALTER TABLE repo_elements
```
phantomii, line 36: Adding published_at with DEFAULT now() gives every existing element the migration timestamp. That makes old packages appear newly published, while the runtime fallback uses created_at when inventory metadata is absent. Please backfill published_at from created_at (or inventory metadata where available) and add migration coverage.

slashburygin: add to migration
session.execute("UPDATE repo_elements SET published_at = created_at")

exordos_core/user_api/repo/api/controllers.py
slashburygin: it's ok now, we don't need filters in this endpoint

exordos_core/user_api/repo/api/controllers.py
slashburygin: done

exordos_core/user_api/repo/api/controllers.py
slashburygin: add StoreControllerMixin

exordos_core/repo/builders/repository.py
slashburygin: ok, added _retire_element

phantomii (approved):

PR #545, opened 2026-07-31
update openapi to 3.1.0

slashburygin:
## Summary by Sourcery

Update OpenAPI specifications and test harness to support both 3.0.3 and 3.1.0 versions, enriching metadata and expanding VM-related schemas and endpoints across user, boot, status, and orchestration APIs.

Enhancements:
- Upgrade boot, status, and orchestration OpenAPI documents from 3.0.3 to 3.1.0 with JSON Schema dialect and richer info metadata (license, contact, terms of service).
- Refocus core schemas and paths around VM, Port, IP address, Tag, and VM definition/filter resources, replacing legacy UniversalAgent and related models in the boot, status, and orch specs.
- Add parallel OpenAPI user specifications for both 3.0.3 and 3.1.0, sharing the same functional-test VM/Port/IP/Tag model surface.
- Extend API surface with additional VM action endpoints (power, power state, power on/off, IP address retrieval) and list/filter endpoints such as VMDefSort, VMNoProcessFilters, and VMNoSort collections.

Tests:
- Adjust manual OpenAPI test to fetch specifications for both 3.0.3 and 3.1.0 versions and to generate versioned user spec files while targeting 3.1.0 for boot, orch, and status APIs.


docs/openapi/openapi_user_3.1.0.yaml
```diff
@@ -0,0 +1,3022 @@
...
+        '200':
+          description: OpenApi specification
+          content:
+            application/json:
+              schema:
+                type: object
+                properties: {}
+        default:
+          $ref: '#/components/responses/Error'
+      operationId: Get_OpenApi_specification
```
phantomii, line 1467: This operationId is duplicated by the 3.0.3 specification endpoint in the same document. Please include the OpenAPI version in the operationId (for example, `Get_OpenApi_specification_3_1_0`) so client generators can consume the spec.

docs/openapi/openapi_boot.yaml
```diff
@@ -54,1096 +63,2403 @@ components:
...
+      summary: Create NotImplementedMethodsRoute NotImplementedMethodsController
+      tags: []
+      parameters: []
+      responses:
+        '201':
+          description: Unknown_Create
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Unknown_Create'
```
phantomii, line 1524: `Unknown_Create` is referenced here, but no matching schema is emitted under `components.schemas`. This makes the generated OpenAPI document invalid and prevents client generation. Please either emit the schema or omit this placeholder route from the published specification.

docs/openapi/openapi_boot.yaml
```diff
@@ -1,8 +1,17 @@
-openapi: 3.0.3
+openapi: 3.1.0
+jsonSchemaDialect: https://spec.openapis.org/oas/3.1/dialect/base
 info:
-  title: Exordos Core v1 Boot API
+  title: REST API Microservice
```
phantomii, line 4: This file is the published Boot API reference, but this update replaces its service metadata and route surface with the functional-test fixture: it now advertises `REST API Microservice`, localhost, and VM test routes that `boot_api` does not expose. Please generate these docs from the actual service contract, or keep the test fixture output out of `docs/openapi`.

docs/openapi/openapi_user_3.1.0.yaml
```diff
@@ -0,0 +1,3022 @@
...
+      type: object
+      required:
+        - status
+        - json
+      properties:
+        status:
+          type: integer
+          description: The HTTP status code
+          minimum: 400
+          exclusiveMaximum: true
```
phantomii, line 36: This document is now declared as OpenAPI 3.1, where `exclusiveMaximum` must be a numeric boundary rather than a boolean. The current `true` value makes the schema invalid for 3.1 tooling. Please remove it or express the intended numeric constraint with the 3.1 syntax.

exordos_core/tests/manual/openapi/test_openapi.py
```diff
@@ -43,19 +47,20 @@ def test_user_openapi(
     ):
         # User API
         client = user_api_client(auth_user_admin)
-        url = f"{user_api.get_endpoint()}{SPECIFICATIONS_PATH}"
-        response = client.get(url, timeout=30)
-        assert response.status_code == 200
+        for version in OPENAPI_VERSIONS:
+            url = f"{user_api.get_endpoint()}{SPECIFICATIONS_PATH}{version}"
+            response = client.get(url, timeout=30)
+            assert response.status_code == 200
```
phantomii, line 53: A 200 response only proves that the endpoint is reachable; the document is then written without any OpenAPI validation. That allowed the invalid specs in this change to be committed. Please validate each generated document before writing it, including unique operationIds and resolvable component references.

docs/openapi/openapi_user_3.1.0.yaml
```diff
@@ -0,0 +1,3022 @@
...
+    variables:
+      version:
+        enum:
+          - v1
+        default: v1
+tags:
+  - name: functional-test
+    description: Just functional tests
+    externalDocs:
+      url: https://https://functional.tests/docs/
```
phantomii, line 3005: The generated `externalDocs` URL has the scheme twice, so it resolves to an invalid destination. This same malformed value is present in all generated specs; please correct the source metadata before publishing the documents.

docs/openapi/openapi_user_3.1.0.yaml
```diff
@@ -0,0 +1,3022 @@
...
+          type: string
+          example: 00000000-0000-0000-0000-000000000000
+          format: uuid
+          readOnly: true
+        state:
+          type: string
+          minLength: 0
+          maxLength: 10
+          example: any_string
+          default: off
```
phantomii, line 79: `state` is declared as a string, but unquoted `off` is parsed by YAML consumers as boolean `false`. This makes the default type-invalid throughout the generated spec. Please emit it as a quoted string (`"off"`) in every affected schema and parameter.

phantomii (changes requested):
Changes requested: the generated OpenAPI documents need to be corrected and validated before merge. See the inline review comments for the required fixes.

PR #552, opened 2026-08-04
add user to project

slashburygin:
## Summary by Sourcery

Enable authorized project members and users with the appropriate permissions to add users to projects with assigned roles.

New Features:
- Add an IAM action for assigning an existing user to a project with a selected role using UUID, email, or username lookup.
- Expose the new project membership permission and document the action in the OpenAPI specification.

Bug Fixes:
- Restrict project read permission checks to direct GET requests while allowing project action handlers to perform their own authorization.

Enhancements:
- Record an audit event when a user is added to a project and avoid creating duplicate role bindings.

Tests:
- Add functional coverage for successful project membership assignment, permission enforcement, invalid users and roles, and owner-role access.

Chores:
- Add an example IAM manifest containing organization, project, user, and role bindings.

### DIFF
--- a/exordos_core/user_api/iam/api/controllers.py
+++ b/exordos_core/user_api/iam/api/controllers.py
@@ -484,16 +484,28 @@ class ProjectController(controllers.BaseResourceControllerPaginated, EnforceMixi
         return models.Project.list_my()
 
     def get(self, uuid, **kwargs):
+        # When active_method is GET, this is a direct resource GET request
+        # and we should enforce permissions. When active_method is None,
+        # get() is called from get_resource_by_uuid() for an action, so
+        # we skip the permission check — the action itself is responsible
+        # for its own permission checks.
         project = super().get(uuid, **kwargs)
-        filters = {"project": ra_filters.EQ(project)}
-        for _ in models.Project.list_my(filters=filters):
-            return project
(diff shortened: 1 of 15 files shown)


exordos_core/user_api/iam/api/controllers.py
```diff
@@ -519,6 +531,87 @@ def delete(self, uuid):
...
+    def add_user(self, resource, user, role):
+        project = resource
+        project_uuid = str(project.uuid)
+        is_member = False
+        filters = {"project": ra_filters.EQ(project)}
+        for _ in models.Project.list_my(filters=filters):
+            is_member = True
+            break
+        if not (
+            is_member
```
phantomii: Any existing project member passes this check, so a regular member can assign themselves the owner role or another role without the required permission. Please restrict this to project owners or users with the dedicated permission, and add a test that a non-owner member is rejected.

phantomii (changes requested):
Please address the inline comment before merging.

exordos_core/user_api/iam/api/controllers.py
slashburygin: ok, now only user, who is owner in his project, can add users to it

phantomii (approved):
Verified the authorization fix and the full CI run. Approved.

akremenetsky (approved):

PR #553, opened 2026-08-04
fix(compute): list machine's volumes before undefining its domain in delete_machine

VasilyStepanov:
Volume-to-machine attribution is derived from the domain's own XML, so looking it up after domain.undefine() always found nothing - delete_machine( delete_volumes=True) silently left every machine's volumes (and their disks) orphaned. List the volumes first, then undefine and delete as before.

## Summary by Sourcery

Ensure machine volumes are discovered before undefining libvirt domains so they can be deleted during machine teardown.

Bug Fixes:
- Fix deletion of a machine's volumes by listing them before the domain is undefined to avoid leaving orphaned volumes and disks.

Tests:
- Add a regression test verifying that deleting a machine with delete_volumes=True removes its associated libvirt storage volumes.

### DIFF
--- a/exordos_core/compute/pool/drivers/libvirt.py
+++ b/exordos_core/compute/pool/drivers/libvirt.py
@@ -1346,19 +1346,22 @@ class LibvirtPoolDriver(base.AbstractPoolDriver):
             )
             domain = None
 
+        # Volume-to-machine attribution is derived from the domain's own
+        # XML (disk attachments), so the volumes must be listed while the
+        # domain is still defined - once undefined below, `list_volumes`
+        # can no longer find them and cleanup would silently skip them.
+        volumes = self.list_volumes(machine) if delete_volumes else ()
+
         if domain is not None:
             # Remove the libvirt domain
             try:
                 domain.destroy()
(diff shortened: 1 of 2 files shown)


exordos_core/compute/pool/drivers/libvirt.py
```diff
@@ -1346,19 +1346,22 @@ def delete_machine(
...
             except libvirt.libvirtError:
                 LOG.debug("The domain is not in the running state")
-            # FIXME(akremenetsky): Actully we should undefine the
-            # domain before volume deletion
             domain.undefine()
 
-        if delete_volumes:
-            for volume in self.list_volumes(machine):
-                self.delete_volume(volume)
+        for volume in volumes:
```
slashburygin, line 1363: why you skip check if delete_volumes?

VasilyStepanov: This check has been moved to the line 1353:
```python
volumes = self.list_volumes(machine) if delete_volumes else ()
```

phantomii (approved):

akremenetsky (approved):

PR #555, opened 2026-08-04
add core_https

slashburygin:
## Summary by Sourcery

Add a dedicated core HTTPS manifest to provision HTTPS vhosts and routing for Exordos Core APIs using wildcard certificates and IAM defaults.

New Features:
- Introduce a core_https manifest defining HTTPS vhost, routes, and certificates for Core API traffic.
- Expose new load balancer backend pool and IAM client variable exports required by HTTPS routing.

### DIFF
--- a/exordos/manifests/core_https.yaml.j2
+++ b/exordos/manifests/core_https.yaml.j2
@@ -1,0 +1,142 @@
+name: "core_https"
+description: "https resources for Exordos Core element"
+schema_version: 1
+version: "{{ version }}"
+api_version: "v1"
+
+imports:
+  core_lb:
+    element: "$core"
+    kind: "resource"
+    link: "$core.network.lb.$core_lb"
+  core_lb_core_backend_http:
+    element: "$core"
+    kind: "resource"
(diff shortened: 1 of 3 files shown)


exordos/manifests/core_https.yaml.j2
```diff
@@ -0,0 +1,142 @@
...
+      project_id: "12345678-c625-4fee-81d5-f691897b8142"
+      parent: $core_https.imports.$core_lb.vhosts.$core_lb_core_https:uuid
+      condition:
+        kind: prefix
+        value: /api/core/v1/iam/clients/
+        allowed_ips:
+          - 0.0.0.0/0
+        actions:
+          - kind: backend
+            pool: $core_https.imports.$core_lb.backend_pools.$core_https.imports.$core_lb_core_backend_http:uuid
```
phantomii, line 93: This route goes directly to the Core backend, while the existing HTTP route uses the IAM cache backend. That means HTTPS introspection and JWKS requests bypass the cache. Could we use `core_lb_iam_cache_backend_http` here as well?

PR #580, opened 2026-08-17
Fixes before SDN

gmelikov:
## Summary by Sourcery

Harden provisioning and networking behavior ahead of SDN integration by fixing DHCP generation, aligning border resource contracts, isolating boot API tests, and tightening build and manifest configuration.

Bug Fixes:
- Prevent invalid empty DNS options from disabling DHCP configuration across all subnets.
- Remove unsupported border routing data from control-plane resources to keep them compatible with the data plane.
- Run boot API functional tests in an isolated service process to avoid cross-test application and database-engine interference.
- Prevent test user fixtures from mutating shared model state.

Enhancements:
- Align manifest examples with the core compute node reference.
- Exclude agent scaffolding from build inputs to avoid filling provisioning storage.
- Add contract coverage for border control-plane and data-plane resource compatibility.

Documentation:
- Update English and Russian manifest examples to use the core compute node reference.

Tests:
- Add DHCP configuration regression tests and border resource compatibility checks.
- Refactor boot API functional tests to exercise the service over HTTP in a separate process.

### DIFF
--- a/docs/em/manifest.md
+++ b/docs/em/manifest.md
@@ -84,7 +84,7 @@ resources:
       group: "ubuntu"
       target:
         kind: "node"
-        node: "$my_app.compute.nodes.$app_node:uuid"
+        node: "$core.compute.nodes.$app_node:uuid"
       service_type:
         kind: "simple"
         count: 1
--- a/docs/em/manifest.ru.md
+++ b/docs/em/manifest.ru.md
@@ -84,7 +84,7 @@ resources:
       group: "ubuntu"
       target:
         kind: "node"
-        node: "$my_app.compute.nodes.$app_node:uuid"
+        node: "$core.compute.nodes.$app_node:uuid"
       service_type:
         kind: "simple"
         count: 1
--- a/exordos/exordos.yaml
+++ b/exordos/exordos.yaml
@@ -16,6 +16,11 @@ build:
         - build
         - cover
         - output
+        # Agent scaffolding, and git worktrees are kept under it -- whole
+        # checkouts with virtualenvs of their own, which the names above do
+        # not reach because they only match at the top level. A build that
+        # meets one fills the provisioning disk and dies in the upload.
+        - .claude
 
     # Binary artifacts. Kernel, initrd, boot loaders.
     - dst: /opt/exordos_core/artifacts/undionly.kpxe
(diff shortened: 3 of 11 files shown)


exordos_core/repo/drivers/base.py
```diff
@@ -49,8 +49,8 @@ def get_element(
     ) -> tuple[repo_models.RepoElement, tp.Collection[repo_models.RepoArtifact]]:
         """Get repository element by name and version."""
 
-    def can_upload_element(self, name: str, version: str) -> bool:
```
akremenetsky, line 52: Why we need to change interface?

exordos_core/compute/builders/pool.py
```diff
@@ -300,10 +300,21 @@ def _actualize_machine_derivatives_on_create_update(
             # Don't swith boot mode for the core set as the update procedure
             # is performed by guest machine driver.
             if not self._is_core_machine(machine):
-                _, guest_actual = machine_guest_pair
+                guest_target, guest_actual = machine_guest_pair
```
akremenetsky: Answered [here](https://github.com/exordos/exordos_core/pull/533#issuecomment-5101624395) let's remove this commit.

exordos_core/cmd/bootstrap.py
```diff
@@ -701,25 +701,40 @@ def _set_defaults_vs(spec: dict[str, tp.Any]):
         {"func": bootstrap_defaults.set_iam_default_client_secret_var, "args": [spec]},
     ]
 
-    # Perform all tasks to set default values until timeout
+    # The tasks are independent of one another — one activates a profile,
```
akremenetsky: Let's discuss. I prefer to keep this snippet simple, we don't need high performance here. Maybe all tasks are independent here but we cannot guarantee new tasks will be independent so the simplest solution that the order make sense and tasks are performed sequentially

exordos_core/repo/drivers/base.py
gmelikov: commit dropped

exordos_core/cmd/bootstrap.py
gmelikov: dropped

exordos_core/compute/builders/pool.py
gmelikov: dropped

akremenetsky (approved):

slashburygin (approved):

PR #581, opened 2026-08-17
feat(dns): let a realm's mirror keep what is not its own

gmelikov:
A managed realm mirrors its zones into the installation above it and
reconciled them by deleting every record it did not have locally --
including the records that installation's own services had put there.
Nothing in a row said whose it was, so "not mine" and "not in my copy"
were the same question.

DNS records carry tags now, and the mirror marks what it writes with
`realm:<realm_uuid>` next to the local record's own tags. It asks the
zone for its marked records only (`?q=tags:"realm:<uuid>"`, off the GIN
index) and removes only what carries the mark. The mark is a label for
a reconciler, not a proof: a client may write it like any other tag.
It is the realm's rather than its IAM user's, since that user is one
per project and can be recreated under the realm.

A row the mirror wrote before it marked its rows is still recognised by
its uuid, which is the local record's: creating it again conflicts and
becomes an update that puts the mark on. A row it wrote unmarked and
has since lost locally cannot be told from anybody else's, and stays.

An ecosystem that cannot read the filter refuses it with a 400, or,
predating the expression, takes `q` for a field and answers 500. Both
mean it has no tags on records either: the whole zone is read, nothing
is marked and nothing removed there. The refusal is remembered for an
hour, so a mirror whose ecosystem is upgraded in place starts marking
on its own. Any other failure is raised rather than retried without
the filter.

Turning on `process_filters` for the record collection changes what
callers see: a query parameter naming no field of the resource is now a
400 where it used to be passed through, and a value is parsed into the
type of the field it names.

A bare `*` is accepted as a record name, the zone's own wildcard, the
way `@` is the zone itself.

### DIFF
--- a/exordos_core/dns_sync/service.py
+++ b/exordos_core/dns_sync/service.py
@@ -21,6 +21,7 @@ import time
 
 import bazooka
 from bazooka import exceptions as bazooka_exc
+from gcl_iam import tokens as iam_tokens
 from gcl_looper.services import basic
 from requests import auth as requests_auth
 from restalchemy.common import contexts
@@ -36,6 +37,14 @@ DNS_SYNC_TIMEOUT = 30
 DNS_SYNC_POOL_SIZE = 1
 FULL_SYNC_INTERVAL = 60
 
+# What an ecosystem that cannot read the filter expression answers with.
+# The current one rejects it as a bad request; one that predates the
+# expression takes `q` for an ordinary field filter, fails to find a field
+# by that name and answers 500. Any other status -- a gateway that is
+# briefly down, a token that stopped working -- says nothing about the
+# filter, so it is not remembered as a refusal and not retried without one.
+FILTER_UNSUPPORTED_STATUSES = frozenset((400, 500))
+
 
 class DNSSyncService(basic.BasicService):
     """Periodically syncs local DNS records to the ecosystem."""
@@ -49,10 +58,12 @@ class DNSSyncService(basic.BasicService):
         self._last_full_sync_at = FULL_SYNC_INTERVAL + 1  # sync immediately on start
         self._last_fast_sync_dt = None
         self._owner_tag = None
-        # Whom the cached owner was asked of: an endpoint or a token
-        # that changes is a different subject to be, and the tag decides
-        # what gets removed, so it is re-asked rather than carried over.
-        self._owner_source = None
+        # The token the cached owner was read out of. A token that
+        # changes is a different subject to be, and the tag decides what
+        # gets removed, so it is re-read rather than carried over. The
+        # cache is not there to save the reading, which is local and
+        # cheap, but to keep the log to one line per token.
+        self._owner_token = None
         # The endpoints that refused the filter expression, so a mirror
         # talking to an older one asks for it only the first time -- and
         # one that moves to a newer one is not stuck with the answer the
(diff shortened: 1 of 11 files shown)


exordos_core/repo/dm/models.py
```diff
@@ -314,9 +315,25 @@ def upload(
         driver = self.load_driver()
 
         # Check if driver supports upload
-        if not driver.can_upload_element(element_name, element_version):
+        if not driver.can_upload_element():
```
akremenetsky: Seems it's not from this PR.

exordos_core/repo/drivers/base.py
```diff
@@ -49,8 +49,8 @@ def get_element(
     ) -> tuple[repo_models.RepoElement, tp.Collection[repo_models.RepoArtifact]]:
         """Get repository element by name and version."""
 
-    def can_upload_element(self, name: str, version: str) -> bool:
-        """Check if element can be uploaded to repository."""
+    def can_upload_element(self) -> bool:
```
akremenetsky: Let's keep the interface as is.

exordos_core/dns_sync/service.py
```diff
@@ -88,6 +92,30 @@ def _eco_get_realm(self, endpoint, realm_uuid, auth):
...
+        """The tag the ecosystem stamps on the records this mirror writes.
+
+        Asked once, of the installation that does the stamping, so the
+        answer cannot drift from what it actually wrote. Without it the
+        mirror still creates and updates -- it only stops removing, which
+        is the safe half to lose.
+        """
+        if self._owner_tag is not None:
+            return self._owner_tag
+        url = f"{endpoint}/api/core/v1/iam/clients/{c.ZERO_UUID}/actions/introspect"
```
akremenetsky: The default client UUID isn't `c.ZERO_UUID` for all installation. Can we use `default` for the default client in the url?

exordos_core/user_api/api/controllers.py
```diff
@@ -14,7 +14,41 @@
...
 from restalchemy.api import controllers
+from restalchemy.common import contexts
+
+from exordos_core.common import constants as c
+from exordos_core.common.dm import models as common_models
+
+LOG = logging.getLogger(__name__)
+
+
+class OwnedTagsControllerMixin:
```
akremenetsky: Let's move this mixin to the common models.

exordos_core/dns_sync/service.py
```diff
@@ -88,6 +98,34 @@ def _eco_get_realm(self, endpoint, realm_uuid, auth):
...
+        try:
+            info = self._client.get(url, headers=headers).json()
+        except Exception:
+            LOG.exception("Failed to ask the ecosystem who this realm is")
+            return None
+        user_uuid = ((info or {}).get("user_info") or {}).get("uuid")
+        if not user_uuid:
+            LOG.warning("Introspection carries no subject; nothing will be removed")
+            return None
+        self._owner_tag = c.owner_user_tag(user_uuid)
```
phantomii: If the configured credentials later introspect as a different IAM user, existing records retain the previous owner tag and disappear from the filtered listing. Recreating them will conflict, while the update keeps the old tag, so the mirror cannot converge. Do we need a migration or handover path for a changed service identity?

phantomii: Can we confirm the ownership invariant here: is this IAM user dedicated to this mirror, with no other DNS records in the same zone? The cleanup decision is based only on this user tag, so any record created by the same identity is treated as belonging to the mirror.

exordos_core/dns_sync/service.py
```diff
@@ -104,20 +142,54 @@ def _eco_create_domain(self, endpoint, headers, name):
...
-            json=record_data,
-            headers=headers,
-        )
+        try:
+            self._client.post(
+                url,
+                json=record_data,
+                headers=headers,
+            )
+        except bazooka_exc.ConflictError:
```
phantomii, line 217: When an untagged legacy record conflicts here, the PUT preserves its lack of an owner tag. The next filtered sync will not see it again, so it will repeat POST → conflict → PUT indefinitely. Could we either claim the record during this transition or keep it visible to the mirror after the fallback?

exordos_core/dns_sync/service.py
```diff
@@ -353,9 +431,21 @@ def _full_sync_domain(self, domain, endpoint, headers, eco_domain_uuid):
...
 
-        # Delete extra
+        # Delete extra -- but only what this mirror put there. A zone in
+        # the ecosystem is not this realm's alone: the ecosystem publishes
+        # the realm's own ingress records into it, and an operator may add
+        # more. Anything not carrying this mirror's owner tag, including
+        # anything created before tags existed, is somebody else's row.
         for uid in eco_uuids - local_uuids:
             eco_rec = eco_map[uid]
+            if not owner_tag or owner_tag not in (eco_rec.get("tags") or []):
```
phantomii, line 472: After this migration, every pre-existing record is untagged and is therefore never removed by the mirror, even after it disappears locally. Is the permanent loss of automatic cleanup for legacy records intended, or do we need a controlled transition/cleanup path?

slashburygin (approved):

exordos_core/dns_sync/service.py
```diff
@@ -353,9 +462,21 @@ def _full_sync_domain(self, domain, endpoint, headers, eco_domain_uuid):
...
```
phantomii, line 472: The ownership key still is not a stable mirror identity. If the access token changes to another IAM subject, the filtered read stops returning the previous subject’s rows, while updates preserve their old owner tag, so the new mirror cannot converge or delete them. Conversely, any unrelated DNS record created with the same IAM subject passes this deletion guard and can be removed when it is absent locally. Please scope ownership to a stable, mirror-specific ID with an explicit handover path, or enforce a dedicated immutable service identity.

PR #595, opened 2026-08-21
feat(em): add $self references to manifest values

slashburygin:
An element gets its uuid at install time, so a manifest author cannot write it down. That blocks binding a `$core.vs.variables` profile setter to the element the variable belongs to, since the setter needs that uuid.

- Resolve `$self:<field>` to a field of the owning element: uuid, name, version, status
- Support it both as a whole value and inside `f"..."` templates
- Raise on an unknown field instead of rendering it empty
- Add unit tests and an example variable bound to its own element

Claude-Session: https://claude.ai/code/session_01WfXnU21LE3UbsKGUXNqnzK

## Summary by Sourcery

Enable manifest resources to reference fields of their owning element, including install-time UUIDs.

New Features:
- Support `$self:<field>` references to resolve an owning element’s uuid, name, version, or status in manifest values and inline templates.

Bug Fixes:
- Reject missing or unsupported `$self` fields with validation errors instead of rendering invalid references as empty values.

Documentation:
- Document `$self` references and their supported fields in the core developer guides.

Tests:
- Add unit coverage for direct and template-based `$self` resolution and validation errors.

Chores:
- Add an example variable whose profile setter targets its owning element via `$self:uuid`.


exordos_core/elements/dm/models.py
```diff
@@ -692,12 +694,33 @@ def get_actual_state_safe(self):
...
+    def _is_self_link(self, link_string):
+        return link_string.partition(":")[0] == self.__self_link__
+
+    def _render_self_value(self, link_string):
+        """Resolve `$self:<field>` to a field of the owning element.
+
+        An element gets its uuid at install time, so a manifest author has no
+        way to spell it out; `$self` lets a resource point at its own element.
+        """
+        field = link_string.partition(":")[2]
```
akremenetsky, line 707: For instance, we have such snippet
```yaml
foo:
  element: $self
```

It leads to the following error -> ` unsupported field `` `.  Let's use more clear error message in this way.

akremenetsky (approved):

exordos_core/elements/dm/models.py
slashburygin: fixed

VasilyStepanov (approved):

PR #611, opened 2026-08-27
scope store endpoints by project

slashburygin:
## Summary by Sourcery

Scope repository store endpoints by project while preserving access to the shared catalogue.

Bug Fixes:
- Scope repository store element reads and related actions to the caller’s project and the shared catalogue, with unscoped and anonymous callers limited to the shared catalogue.

Enhancements:
- Remove the obsolete store-element read permission and role binding now that store access is enforced through project scoping.

Tests:
- Update repository API coverage for project-specific, shared-catalogue, and anonymous store element visibility.

### DIFF
--- a/exordos_core/user_api/repo/api/controllers.py
+++ b/exordos_core/user_api/repo/api/controllers.py
@@ -89,10 +89,7 @@ class RepositoryController(
 
 
 class StoreControllerMixin(iam_controllers.PolicyBasedWithoutProjectController):
-    """Policy and project scoping shared by the store endpoints."""
-
-    __policy_service_name__ = "repo"
-    __policy_name__ = "store_element"
+    """project scoping shared by the store endpoints."""
 
     def _project_filters(self):
         """Restrict a read to the projects the caller may see.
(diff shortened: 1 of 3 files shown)


exordos_core/user_api/repo/api/controllers.py
```diff
@@ -183,13 +176,19 @@ class StoreElementController(
...
     def get(self, uuid, **kwargs):
-        # Scope the lookup so an action cannot reach another project.
         kwargs.update(self._project_filters())
-        return super().get(uuid=uuid, **kwargs)
+        return controllers.BaseResourceControllerPaginated.get(
+            self, uuid=uuid, **kwargs
+        )
+
+    def filter(self, filters, order_by=None):
+        filters.update(self._project_filters())
```
akremenetsky, line 185: So if there is no `_ctx_project_id`, an user gets all `RepoElement` even though in private projects? Is it possible this user is not an admin?

slashburygin: A caller with no project — a token issued without a project: scope, or an anonymous request, which AnonDriver also reports as project-less — now reads the shared catalogue only, instead of every project

akremenetsky (approved):

phantomii (approved):

PR #613, opened 2026-08-28
Add manifest download actions

slashburygin:
## Summary

Adds a `download` action that hands back a manifest as a YAML file, for both
installed manifests (EM) and catalogue elements (repo store). Until now the
manifest was only readable as part of a JSON resource, so getting the
original document back out meant reassembling it by hand.

## Changes

- `GET /v1/em/manifests/<uuid>/actions/download` returns the stored manifest
  as `application/yaml` with `Content-Disposition: attachment`. The fields the
  API keeps around the document (`uuid`, `status`, `project_id`, `created_at`,
  `updated_at`) are left out, so the file round-trips back through
  `POST /v1/em/manifests`.
- `GET /v1/repo/store/elements/<uuid>/actions/download` does the same for
  `RepoElement.manifest`. It enforces `repo.store_element.read` like
  `stable_versions`, and actualizes the element first when a lazy repository
  has not fetched the manifest yet — the check `RepoElementController.get`
  already makes.
- Both build the response through a shared helper in
  `exordos_core/common/api/manifests.py`; the filename is sanitized to
  `[A-Za-z0-9._-]` so an element name cannot break the header.
- Both controllers get `__packer__ = packers.JSONPackerPreEncoded` so the
  encoded YAML body passes through untouched while every other response stays
  JSON — the pattern `boot_api` already uses.

No new policy entries: `em.manifest.*` permissions do not exist and the other
manifest actions do not enforce, so `download` follows suit; the store action
reuses the existing `repo.store_element.read`. `ec-manifest-schema` produces
no change to `full_spec.yaml`.

## Verification

- `pytest exordos_core/tests/functional -n 10` — 534 passed, 3 skipped
- `pytest exordos_core/tests/unit -n 10` — 261 passed
- `ruff check` and `ruff format --check` on the touched files — clean
- `ec-manifest-schema` — `full_spec.yaml` unchanged

## Summary by Sourcery

Enable users to download installed and catalogue manifests as reusable YAML documents.

New Features:
- Add download actions for installed manifests and repository elements that return their manifests as downloadable YAML files.

Enhancements:
- Centralize YAML manifest response generation with sanitized filenames and appropriate download headers.
- Preserve JSON responses for existing controller actions while allowing pre-encoded YAML downloads.
- Actualize lazy repository elements before downloading their manifests and enforce repository read access.

Tests:
- Add functional coverage for manifest download content, headers, filenames, and omitted service metadata.
- Add functional coverage for repository element manifest downloads.

### DIFF
--- a/exordos_core/user_api/em/api/controllers.py
+++ b/exordos_core/user_api/em/api/controllers.py
@@ -21,6 +21,7 @@ from gcl_iam.api import controllers as iam_controllers
 from restalchemy.api import actions
 from restalchemy.api import constants
 from restalchemy.api import controllers
+from restalchemy.api import packers
 from restalchemy.api import resources
 from restalchemy.common import exceptions as ra_e
 from restalchemy.dm import filters as dm_filters
@@ -28,8 +29,20 @@ from restalchemy.openapi import constants as oa_c
 from restalchemy.openapi import utils
 import webob
 
+from exordos_core.common.api import manifests
 from exordos_core.elements.dm import models
 from exordos_core.vs.dm import models as vs_models
+
+
+# Fields the API keeps around the manifest document. The downloaded file is
+# the manifest as it was uploaded, so they are left out of it.
+MANIFEST_SERVICE_FIELDS = (
+    "uuid",
+    "status",
+    "project_id",
+    "created_at",
+    "updated_at",
+)
 
 
 class ElementHasNoProfileError(ra_e.ValidationErrorException):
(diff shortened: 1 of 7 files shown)


exordos_core/user_api/em/api/controllers.py
```diff
@@ -101,6 +117,14 @@ def uninstall(self, resource):
     def validate(self, resource):
         return resource.validate_schema_base()
 
+    @actions.get
+    def download(self, resource):
```
akremenetsky, line 121: IAM permissions?

PR #620, opened 2026-09-03
Reduce memory usage of the repo proxy general service

slashburygin:
## Summary

The repo proxy general service (`exordos-repo-proxy-gservice`) held full ORM
records alive in paths that only ever read a UUID off them, so its peak
footprint scaled with manifest payload size instead of element count. This
trims the retention in the two hot paths without changing behaviour.

The remaining ~75 MiB is import-time cost (`launchpad`, `restalchemy`,
`gcl_sdk`, `elements.dm.models`), shared by all three services the launchpad
hosts and not reducible without dropping dependencies.

## Changes

- `RepoEmBackendClient._list` runs on every agent iteration (`iter_min_period=3`)
  and loaded every EM `Manifest` *and* every EM `Element`, holding both while
  building the result. `Element.manifest` and `Element.profile` are declared
  `prefetch=True`, which restalchemy compiles into a JOIN, so each element
  dragged in a second full copy of `requirements`, `resources`, `exports`,
  `imports` and `openapi_spec` — all to read `em_element.uuid`. The element
  query now yields a `(name, version) -> uuid` map and runs first, so those
  objects are released before the manifests are read.
- `_make_installed_manifest` takes the element UUID rather than the model.
  Manifest dedup by `(name, version)` is kept deliberately: `em_manifests` has
  no unique constraint on that pair, unlike `em_elements`.
- `_refresh_repository` keyed every element of a repository by its full record;
  those carry the `manifest`, `specification` and `inventory` dicts and are
  needed only for the few elements being deleted, which are now re-fetched in a
  single `ra_filters.In` query.
- `_refresh_repository` also materialised the `iter_elements_in_inventory`
  generator and kept every entry. In COPY mode `actualize_element` downloads the
  full payload into each one, so all of them accumulated for the whole refresh;
  entries are now dropped from the mapping as they are saved.

What this does and does not change: `get_all()` still materialises every full
record for the length of that call, and there is no way around that with the
ORM as it stands — both `ObjectCollection._get_all` and `_query` end in
`restore_from_storage(**params)`, with no column projection available. What
goes away is the *concurrent* retention: the full-record maps used to stay
alive across the inventory loop, where in COPY mode `actualize_element` builds
a second full payload set on top of them, so peak was roughly 2x the element
set; it is now ~1x, the transient `get_all` spike.

Measured on a stand-in with 300 elements and realistically sized manifests, the
`_list` element mapping retains 13.92 MiB before and 0.07 MiB after. That figure
is the retained map, not the transient peak. The stand-in omits profiles, so the
real saving is somewhat larger.

Not addressed: the `NOTE(akremenetsky)` in `_list` proposes a joined view to
replace the separate manifest/element queries, which would also cut the
remaining manifest materialisation. That needs a migration and is left alone
here.

## Changes to tests

Three unit tests passed a mock element object into `_make_installed_manifest`;
they now pass `.uuid`.

## Verification

- `ruff check .` — clean
- `pytest exordos_core/tests/unit -n 10` — 265 passed
- `pytest exordos_core/tests/functional -n 10` — 540 passed, 3 skipped
  (includes `test_refresh_repository_removes_old_elements`,
  `..._keeps_installed_elements`, `..._updates_latest`, and both COPY and LAZY
  iteration modes)
- `mypy` — no new errors; diffed against the base, the one pre-existing
  `_list(**kwargs)` annotation error just shifts a line

## Summary by Sourcery

Reduce peak memory usage in repository synchronization paths without changing repository behavior.

Enhancements:
- Reduce repository proxy memory retention by storing only element UUIDs during manifest and repository refresh operations.
- Release downloaded inventory elements as they are processed and re-fetch only stale records needed for deletion.

Tests:
- Update unit tests for UUID-based installed manifest construction.

### DIFF
--- a/exordos_core/repo/builders/repository.py
+++ b/exordos_core/repo/builders/repository.py
@@ -22,6 +22,7 @@ from gcl_looper.services.oslo import base as oslo_base
 from gcl_sdk.agents.universal.dm import models as ua_models
 from gcl_sdk.agents.universal.services import builder as sdk_builder
 from packaging import version as packaging_version
+from restalchemy.dm import filters as ra_filters
 
 from exordos_core.repo.dm import models
 
@@ -113,9 +114,11 @@ class RepoProxyBuilderService(
 
     def _refresh_repository(self, instance: models.Repository) -> None:
         """Actualize elements in the repository."""
-        # Get existing elements for this repository
-        existing_elements = {
-            (elem.name, elem.version): elem
+        # Get existing elements for this repository. Only the UUIDs are kept:
+        # the full records carry the manifest, specification and inventory
+        # dicts, and are needed only for the few elements to be deleted below.
+        existing_uuids = {
+            (elem.name, elem.version): elem.uuid
             for elem in models.RepoElement.objects.get_all(
                 filters={"repository": instance.uuid},
             )
@@ -125,9 +128,12 @@ class RepoProxyBuilderService(
         inventory_elements = {
             (e.name, e.version): e for e in instance.iter_elements_in_inventory()
         }
+        inventory_keys = frozenset(inventory_elements)
         new_elements_count = 0
-        for key in inventory_elements.keys() - existing_elements.keys():
-            element = inventory_elements[key]
+        for key in inventory_keys - existing_uuids.keys():
+            # Drop the element from the mapping once handled: in COPY mode it
+            # holds the full downloaded manifest, specification and inventory.
+            element = inventory_elements.pop(key)
 
             # In COPY mode, actualize element to download full data
             if instance.sync_mode == models.SyncMode.COPY:
@@ -147,8 +153,10 @@ class RepoProxyBuilderService(
         # Remove old elements that are no longer in inventory,
         # but keep installed ones
         deleted_elements_count = 0
-        for key in existing_elements.keys() - inventory_elements.keys():
-            element = existing_elements[key]
+        for key in existing_uuids.keys() - inventory_keys:
+            element = models.RepoElement.objects.get_one(
+                filters={"uuid": ra_filters.EQ(existing_uuids[key])},
+            )
 
             # Skip deletion if element is installed
             if (
--- a/exordos_core/repo/agents/universal/drivers/repo_element.py
+++ b/exordos_core/repo/agents/universal/drivers/repo_element.py
@@ -17,6 +17,7 @@
 import logging
 import os
 import typing as tp
+import uuid as sys_uuid
 
 from gcl_sdk.agents.universal.clients.backend import db as client
 from gcl_sdk.agents.universal.clients.backend import exceptions as backend_exc
@@ -60,7 +61,7 @@ class RepoEmBackendClient(client.DatabaseBackendClient):
     def _make_installed_manifest(
         self,
         em_manifest: em_models.Manifest,
-        em_element: em_models.Element | None,
+        em_element_uuid: sys_uuid.UUID | None,
     ) -> re_builder.InstalledManifest:
         """Build an InstalledManifest from an EM manifest and its counterparts.
 
@@ -80,7 +81,7 @@ class RepoEmBackendClient(client.DatabaseBackendClient):
             name=em_manifest.name,
             version=em_manifest.version,
             description=em_manifest.description,
-            element=em_element.uuid if em_element is not None else None,
+            element=em_element_uuid,
             status=models.RepoElementStatus.ACTIVE.value,
             project_id=em_manifest.project_id,
         )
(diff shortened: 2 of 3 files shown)


gmelikov (approved):

exordos_core/repo/builders/repository.py
```diff
@@ -147,8 +153,10 @@ def _refresh_repository(self, instance: models.Repository) -> None:
         # Remove old elements that are no longer in inventory,
         # but keep installed ones
         deleted_elements_count = 0
-        for key in existing_elements.keys() - inventory_elements.keys():
-            element = existing_elements[key]
+        for key in existing_uuids.keys() - inventory_keys:
+            element = models.RepoElement.objects.get_one(
```
akremenetsky: The change reduces simultaneous retention of full ORM collections, but introduces an unbounded N+1 query path for stale elements. A repository losing K inventory entries now performs K additional full-row SELECTs. Please fetch stale rows in one query or bounded batches. Also note that the initial get_all() still materializes every heavy payload, so the commit’s peak-memory claim requires a lightweight projection/view to be true.

exordos_core/repo/builders/repository.py
```diff
@@ -113,9 +114,11 @@ def post_create_instance_resource(
...
         """Actualize elements in the repository."""
-        # Get existing elements for this repository
-        existing_elements = {
-            (elem.name, elem.version): elem
+        # Get existing elements for this repository. Only the UUIDs are kept:
+        # the full records carry the manifest, specification and inventory
+        # dicts, and are needed only for the few elements to be deleted below.
+        existing_uuids = {
+            (elem.name, elem.version): elem.uuid
             for elem in models.RepoElement.objects.get_all(
```
akremenetsky, line 122: This does not significantly reduce peak memory usage. `get_all()` still fetches and materializes every full `RepoElement`, including its potentially large `manifest`, `specification`, and `inventory` fields. The comprehension only releases those objects after the complete result has already been loaded, so the main memory peak remains. Given that stale elements are then fetched again one by one, introducing an N+1 query pattern, the practical value of this optimization is questionable. A lightweight projection/view that selects only `name`, `version`, and `uuid` would be needed to provide a meaningful memory improvement.

exordos_core/repo/builders/repository.py
slashburygin: Fixed in b8cf728 — the stale elements are now fetched with a single `ra_filters.In` query instead of a `get_one` per key. No empty-list guard is needed: restalchemy's `In._convert_value` already substitutes `[None]` for an empty list.

Worth noting the loop body is still inherently per-element — each iteration does `element.delete()`, and possibly `_retire_element` or `_set_latest_for_outdated_element` — so this batches the SELECTs, not the round-trips overall.

exordos_core/repo/builders/repository.py
slashburygin: Agreed that `get_all()` still materializes every full record — there's no way around that with the ORM as it stands: both `ObjectCollection._get_all` and `_query` end in `restore_from_storage(**params)`, so there is no column projection available. A real projection needs raw SQL or a view plus a migration, which is the `NOTE(akremenetsky)` case this PR deliberately left alone.

On "the main memory peak remains" — I'd put it differently. Before this change the full-record map stayed alive across the whole inventory loop, and in COPY mode `actualize_element` builds a second full payload set on top of it, so peak was roughly 2x the element set. Now the records are released at the comprehension and peak is ~1x, the transient `get_all` spike. So it halves concurrent retention rather than eliminating the spike. The 13.92 -> 0.07 MiB figure in the description measures the retained map, not that transient; I've reworded it to make that clear.

The N+1 is fixed in b8cf728.

exordy (approved):
### exo-review: 1 low, nothing blocking

Intent: Cut ORM retention in the repo proxy's two hot paths by keeping element UUIDs instead of full records.

#### Outside the diff

- `exordos_core/repo/agents/universal/drivers/repo_element.py:71` **[low]** The docstring promises a repo-element payload substitution the function has never done, and the new signature makes that plainer. **Fix:** Drop the second paragraph. <details><summary>Why</summary>

Nothing but `em_element_uuid` reaches the result; the manifest dict comes from `em_manifest` alone. Pre-existing prose, but the parameter it describes no longer even exists.

</details> 

#### Verify before merge

- The 13.92 -> 0.07 MiB retention figure was not reproduced; rerun the author's stand-in under tracemalloc.
- A repository dropping tens of thousands of entries sends that many uuids in one IN; check the query size if such repos exist.

<details>

**Clean axes:** security — no tenancy, API or log surface; uuids come from the per-repo query; scope — three files, every hunk traces to the memory intent; compat — no migration, API, manifest or target-resource shape touched; reconciliation — delete set and dedup keys unchanged; UNIQUE(repository,name,version) makes the uuid map…; tests — functional refresh tests cover add/remove/retire on a real DB; prod-readiness — no TODO, debug or process comments; conventional commits

**Checks run:**
- read patch + head of both changed files — matches
- grep restalchemy sql/filters.py: In([]) -> [None] — claim holds
- grep migrations: em_elements UNIQUE(name,version), em_manifests none — claim holds
- grep repo_elements UNIQUE(repository,name,version) — uuid map lossless
- grep _make_installed_manifest call sites — in-repo only, no base-class override

</details>

<sub>16 turns · 258s</sub>

PR #623, opened 2026-09-03
feat(compute): schedule disks onto storage pools by speed/ephemeral

VasilyStepanov:
## Summary by Sourcery

Enable storage-aware disk scheduling with speed tiers, ephemeral constraints, and durable pool assignment tracking.

New Features:
- Support disk speed tiers and ephemeral placement requirements in resource specifications.
- Schedule disks across storage pools using speed matching, capacity availability, and tracked pool assignments.
- Reuse existing machine volumes according to their actual storage pool tier while safely skipping unresolved legacy assignments.

Bug Fixes:
- Prevent disk placement from assuming the first storage pool and avoid incorrect capacity accounting when reusing volumes.

Enhancements:
- Persist disk placement attributes and enforce speed and ephemeral fields as immutable after volume creation.

Documentation:
- Document disk speed and ephemeral properties and their storage-pool placement behavior in English and Russian resource guides.
- Update the full API specification with disk, volume, and storage-pool placement fields.

Tests:
- Add coverage for immutable volume placement attributes and storage-pool selection and reuse fallbacks.

Chores:
- Add a database migration for volume speed, ephemeral, and storage-pool tracking fields.


exordos_core/compute/dm/models.py
```diff
@@ -358,6 +359,18 @@ class MachineVolume(
     )
     # TODO(g.melikov): DON'T USE! Should be dropped.
     device_type = properties.property(types.String(max_length=64), default="")
+    speed = properties.property(
```
akremenetsky, line 375: What we are going to do if an user updates the `speed`? Perhaps the target solution is migration to a corresponding storage pool but now we just ignore it. Would it be better to set this paramters as read-only?

exordos_core/compute/scheduler/service.py
```diff
@@ -179,15 +179,47 @@ def _get_pools(
             for p in pools
         )
 
+    def _select_storage_pool(
```
akremenetsky, line 182: Would we like to change signature of the method? If a new parameters is added to the volume, we will need to change the interface of the method too. For instance, we add `cost` parameter to the volume:

```python
def _select_storage_pool(
        self,
        pool: base.MachinePoolBundle,
        speed: str,
        ephemeral: bool,
        size: int,
        cost: str # new parameter
    )
```

All these parameters belong to the volume, so we can pass the volume object instead of individual parameters:
```python
def _select_storage_pool(
        self,
        pool: base.MachinePoolBundle,
        volume: models.MachineVolume
    )
```

exordos_core/compute/scheduler/service.py
```diff
@@ -221,46 +268,78 @@ def _place_volume_into_pool(
...
 
-        # Figure out the best volume
-        volumes.sort(key=lambda v: volume.size - v.size)
-        pool_volume = volumes[0]
-        need_size = volume.size - pool_volume.size
+            if pool_volume.storage_pool:
+                storage_pool = self._find_storage_pool_by_name(
+                    pool, pool_volume.storage_pool
+                )
+            else:
```
akremenetsky: Actually I cannot understand how this branch is possible. The idea of this cycle if to find `hot volume` that fit all requirements and place user payload there. If a volume exist, it means its scheduling procedure has been completed and that means we already known its storage pool.
Perhaps a volume hasn't been completed the scheduling procedure, but we can skip such volumes to follow common and don't do scheduling here.

exordos_core/compute/scheduler/service.py
```diff
@@ -221,46 +268,78 @@ def _place_volume_into_pool(
...
+
+        for pool_volume in exact_volumes + other_volumes:
+            need_size = volume.size - pool_volume.size
 
-        # Figure out the best volume
-        volumes.sort(key=lambda v: volume.size - v.size)
-        pool_volume = volumes[0]
-        need_size = volume.size - pool_volume.size
+            if pool_volume.storage_pool:
+                storage_pool = self._find_storage_pool_by_name(
```
akremenetsky: The `_find_storage_pool_by_name` raises `ValueError` and single broken `pool_volume` can lead to unhandled exception.

exordos_core/compute/pool/dm/models.py
```diff
@@ -138,6 +138,8 @@ def get_resource_target_fields(self) -> tp.Collection[str]:
                 "boot",
                 "label",
                 "device_type",
+                "speed",
+                "ephemeral",
```
akremenetsky, line 142: We need to add `storage_pool` as well

exordos_core/compute/scheduler/service.py
```diff
@@ -179,15 +179,48 @@ def _get_pools(
...
+        """Select a storage pool for `volume`.
+
+        `size` is passed separately (rather than using `volume.size`)
+        since callers also use this for a resize's capacity delta.
+
+        See gcl_sdk's `ua_pool.select_storage_pool` for the matching
+        rules (soft speed/ephemeral match with a capacity fallback).
+        """
+        storage_pool = ua_pool.select_storage_pool(
+            pool.pool.storage_pools, volume.speed, volume.ephemeral, size
```
exordy, line 194: **[critical] A durable disk falls back onto an ephemeral storage pool when no durable pool has room.**

**Mechanism:** gcl_sdk's `select_storage_pool` fallback branch (feat/disk-scheduling, pool.py:392) filters on capacity alone — `ephemeral` is dropped together with `speed`. Both call sites here, and docs/em/resources.md:98, treat it as a hard constraint.

**Scenario:** Pool has `nvme-cache` (ephemeral) and a full durable `bulk`; a node with the default `ephemeral: false` root disk lands on `nvme-cache` and loses its data on host reboot.

**Fix:** Restrict the fallback set to pools with the requested `ephemeral` in gcl_sdk's `select_storage_pool`; both call sites here then inherit it.

exordos_core/compute/builders/pool.py
```diff
@@ -547,11 +568,15 @@ def _can_update_volume(
...
-        storage_pool: ua_pool.AbstractStoragePool = volume.pool.storage_pools[0]
+        storage_pool = self._select_storage_pool(
+            volume.pool, target_volume, target_volume.size - actual_volume.size
+        )
 
         # FIXME(akremenetsky): Does it work correctly?
         # Will every volume refer to own pool object?
         storage_pool.allocate_capacity(target_volume.size - actual_volume.size)
+        if not volume.storage_pool:
+            volume.storage_pool = storage_pool.name
```
exordy, line 579: **[high] A legacy volume with `storage_pool` NULL gets pinned to a guessed pool on its first resize.**

**Mechanism:** The migration leaves `storage_pool` NULL and gives every legacy row the same speed/ephemeral defaults, so the soft match just returns whichever pool has most free space. Commit 1751985 removed this guess from the scheduler as unsafe; here it survives.

**Scenario:** Upgraded pool with `ssd` and `hdd`; a legacy volume physically on `hdd` is resized, the match returns `ssd`, and the delta is charged to `ssd` while `hdd` fills.

**Fix:** Leave `storage_pool` unset when it cannot be determined, or backfill it in the migration to the historical first pool.

docs/em/resources.md
```diff
@@ -65,6 +65,8 @@ disk_spec:
   kind: "root_disk"
   size: 10
   image: "https://repo.exordos.com/exordos-base/1.1.0/exordos-base.raw.zst"
+  speed: "HOT"      # optional: COLD | WARM (default) | HOT
```
exordy, line 68: **[high] The uppercase `COLD`/`WARM`/`HOT` values documented and spec'd here exist in no pushed gcl_sdk.**

**Mechanism:** `ic.DiskSpeed` is `cold`/`warm`/`hot` at feat/disk-scheduling's tip (f80c4ad) and absent from releases up to 3.6.0. `types.Enum` over it therefore rejects `HOT`, and full_spec.yaml's uppercase enums cannot be reproduced by `ec-manifest-schema`.

**Scenario:** A manifest written from this page with `speed: "HOT"` is rejected at validation, and `test_full_spec_is_current` fails against the pinned SDK.

**Fix:** Settle the casing in gcl_sdk first, then regenerate full_spec.yaml and align both doc pages and the migration's CHECK to it.

migrations/0003-volumes-speed-ephemeral-fba1de.py
```diff
@@ -0,0 +1,76 @@
...
+            f"""
+            ALTER TABLE compute_machine_volumes
+                ADD COLUMN speed character varying(16)
+                    DEFAULT 'warm' NOT NULL,
+                ADD COLUMN ephemeral boolean DEFAULT false NOT NULL,
+                ADD COLUMN storage_pool character varying(255),
+                ADD CONSTRAINT compute_machine_volumes_speed_check
+                    CHECK (speed IN {DISK_SPEED_CHECK});
+            """
+        )
```
exordy, line 56: **[medium] `pool_volume`'s target-field set grew but nothing nudges `compute_machine_volumes.updated_at`.**

**Mechanism:** `get_updated_entities` rebuilds `ua_target_resources` only where `compute_machine_volumes.updated_at != tracked_at`, and ALTER TABLE moves neither. Pre-upgrade rows keep sending the pre-`storage_pool` value until some other write touches them.

```suggestion
        )
        session.execute("UPDATE compute_machine_volumes SET updated_at = now();")
```

exordos_core/tests/unit/compute/test_scheduler_volume_placement.py
```diff
@@ -0,0 +1,238 @@
...
+        pool_filters=[],
+        pool_weighters=[],
+        machine_filters=[],
+        machine_weighters=[],
+    )
+
+
+class TestPlaceVolumeIntoPoolFallback:
+    """A full exact-tier candidate must not block reuse of a different,
+    still-fitting candidate (Codex review on PR #623).
```
exordy, line 84: **[low] Test docstrings record who asked for the change on PR #623.**

**Mechanism:** Review attribution says nothing about what the code does and rots at the next edit; the commit bodies already carry it. Same at lines 150 and 182, and test_models.py:54.

**Fix:** State the invariant each test protects and drop the attributions.

exordy (commented):
### exo-review: 1 critical, 2 high, 1 medium, 1 low

Intent: Place disks on storage pools by speed/ephemeral tier and persist the chosen pool, instead of always taking storage_pools[0].

#### Verify before merge

- Upgrade convergence on a stand: after two agent iterations both compat.md queries (target vs actual hash, ua_outdated_tracked_resources_view) must return empty.
- That a durable request actually reaches an ephemeral pool end to end: needs a stand with one ephemeral and one full durable storage pool.

Note: gcl_sdk feat/disk-scheduling (f80c4ad) is the only pushed source of select_storage_pool; no release up to 3.6.0 has it.

Note: Reuse looks unreachable today: nothing in the tree sets MachineVolume.node_volume back to NULL, so pool.volumes is always empty.

Note: Reuse never updates pool_volume.project_id — pre-existing, but the widened fallback makes it easier to hit.

<details>
<summary>Coverage: 6 axes clean · 4 checks run · mode shipped</summary>

**Clean axes:** security — no tenancy, permission or secret surface touched; diff scope — every hunk traces to disk tiering; reconciliation — a broken reuse candidate no longer aborts the node's placement; bugfix evidence — each fix commit names its mechanism and lands a test; prod-readiness — no TODO/debug/WIP leftovers added; hygiene — Apache headers, English, commit format, both doc languages in step

**Checks run:**
- git diff/log over the range — read
- axis-9 greps for WIP and history comments — one hit (finding 5)
- axis-6 greps: target-resource change present, migration nudge absent
- gcl_sdk refs/tags inspected for DiskSpeed and select_storage_pool

</details>

<sub>37 turns · 524s</sub>

PR #627, opened 2026-09-06
Name the personal workspace project after its owner

slashburygin:
## Summary

A newly provisioned personal workspace used to get a project literally named
`default`. It is now named after its owner — `<username>'s project`.

## Changes

- `provision_personal_workspace()` creates the project as
  `f"{self.name}'s project"` (description stays `Default project`)
- Updated the provisioning unit and functional test assertions

Nothing looks a project up by the name `default` — `Project.get_default()`
resolves it through the earliest role binding — so no lookup path changes.
Projects that already exist are renamed out of band, not by this PR.

## Note

`iam_projects.name` is `varchar(128)` while a username may be up to 128
characters, so `<username>'s project` can overflow the column. Worth deciding
whether to widen the column or truncate in `provision_personal_workspace()`.

## Verification

- `python -m pytest exordos_core/tests/unit/user_api/iam/ -q` — 70 passed
- `python -m pytest exordos_core/tests/functional/restapi/iam/test_users.py -k provision -q` — 3 passed
- `ruff format --check` / `ruff check` on the touched files — clean
- Full `tox` suite not run


exordos_core/user_api/iam/dm/models.py
```diff
@@ -465,7 +465,7 @@ def provision_personal_workspace(self):
             role=iam_c.OrganizationRole.OWNER.value,
         ).insert()
         project = Project(
-            name="default",
+            name=f"{self.name}'s project",
```
exordy, line 468: **[high] A username longer than 118 characters makes the personal project name overflow iam_projects.name varchar(128), which rolls back the entire email confirmation.**

**Mechanism:** types.Username(min_length=1, max_length=128) accepts a 128-character name; f"{self.name}'s project" adds 10 characters while iam_projects.name is character varying(128). iam_organizations.name is also 128, so the org and the membership insert succeed and the project insert raises psycopg2 DataError. The whole HTTP request runs inside one transaction (ContextMiddleware.process_request -> ctx.session_manager() in restalchemy/api/middlewares/contexts.py:133), so the exception rolls back resource.confirm_email() too, and every retry takes the identical path.

**Scenario:** A user self-registers with a 119-character username through an IamClient with registration_auto_provision=true, then POSTs .../confirm_email: the response is a 500, email_verified stays false and no workspace exists, permanently. Before this PR the project name was the constant "default", so this could not happen.

**Fix:** Truncate the composed name to the column width so it agrees with the migration's left(..., 128); alternatively widen iam_projects.name, but that needs a migration hunk in this same PR.

```suggestion
            name=f"{self.name}'s project"[:128],
```

migrations/0004-rename-default-projects-9af1ca.py
```diff
@@ -0,0 +1,68 @@
...
+
+    def upgrade(self, session):
+        # iam_projects.name is varchar(128), so a long username is cut off.
+        session.execute(
+            f"""
+            UPDATE public.iam_projects p
+            SET name = left(o.owner || '''s project', 128),
+                updated_at = now()
+            FROM ({PROJECT_OWNERS}) o
+            WHERE p.uuid = o.project AND p.name = 'default';
```
exordy, line 51: **[medium] The migration renames every project literally named "default", including ones users created themselves, not only auto-provisioned personal workspaces.**

**Mechanism:** The only predicate is p.name = 'default'. ProjectController.create calls project.add_owner(models.User.me()) (exordos_core/user_api/iam/api/controllers.py:478), so every API-created project has a project-scoped OWNER role binding and therefore matches the PROJECT_OWNERS join. Nothing restricts the update to the provisioned workspace (project description 'Default project', organization description 'Personal workspace').

**Scenario:** In a shared organization "acme", Bob creates a project named "default" via POST /organizations/{acme}/projects. On upgrade the migration silently renames it to "bob's project". After the upgrade nothing distinguishes it from an auto-provisioned project, so the change cannot be undone selectively.

**Fix:** Scope the upgrade predicate to the provisioned workspace: require p.description = 'Default project' and the owning organization to be a personal workspace. Apply the same predicate to the downgrade (lines 62-63), which otherwise clobbers any project a user deliberately named "<owner>'s project".

```suggestion
            WHERE p.uuid = o.project
              AND p.name = 'default'
              AND p.description = 'Default project'
              AND EXISTS (
                  SELECT 1
                  FROM public.iam_organizations org
                  WHERE org.uuid = p.organization
                    AND org.description = 'Personal workspace'
              );
```

migrations/0004-rename-default-projects-9af1ca.py
```diff
@@ -0,0 +1,68 @@
...
+    def is_manual(self):
+        return False
+
+    def upgrade(self, session):
+        # iam_projects.name is varchar(128), so a long username is cut off.
+        session.execute(
+            f"""
+            UPDATE public.iam_projects p
+            SET name = left(o.owner || '''s project', 128),
+                updated_at = now()
```
exordy, line 49: **[low] now() writes session-local time into a `timestamp without time zone` column, while the application writes UTC.**

**Mechanism:** now() returns timestamptz; assigning it to timestamp(6) without time zone casts through the session's TimeZone GUC. The application writes true UTC instead: UTCDateTimeZ.to_simple_type (restalchemy/dm/types.py:788) emits a naive UTC string that Postgres stores verbatim. The same expression is on line 60 in the downgrade.

**Scenario:** On a Postgres whose timezone GUC is 'Europe/Moscow', every project renamed by this migration gets an updated_at three hours ahead of the UTC values in the rest of the table, so an updated_at-ordered listing or a "changed since" comparison places them wrongly.

**Fix:** Use `updated_at = (now() at time zone 'utc')` in both upgrade and downgrade.

```suggestion
                updated_at = (now() at time zone 'utc')
```

exordy (commented):
### exo-review: 1 high, 1 medium, 1 low: A username longer than 118 characters makes the personal project name overflow iam_proj…

Intent: Name the auto-provisioned personal-workspace project after its owner instead of "default", and migrate the projects that already carry that name.

#### Verify before merge

- The blast radius of the over-broad rename on a real installation: run `SELECT count(*) FROM iam_projects WHERE name = 'default';` before applying, and compare with the number of personal workspaces (`SELECT count(*) FROM iam_organizations WHERE description = 'Personal workspace';`).
- End-to-end registration with a 119-character username (finding 1) — needs a live stand: register through a client with registration_auto_provision=true and POST .../confirm_email; the unit suite mocks Project, so it cannot reach the column limit.

Note: The PR description states that `iam_projects.name` is varchar(128) "while a username may be up to 128 characters"; the iam_users.name column is in fact varchar(256) — the 128 limit comes from the Username DM type (exordos_core/user_api/iam/dm/types.py:23). The conclusion about the overflow is unchanged.

Note: The unit and functional assertions were changed to `f"{user.name}'s project"` rather than a literal. They still fail if the implementation's format changes, so this is not a tautological assertion — just less readable than the literal "testuser's project".

Note: Migration 0003 subclasses the legacy misspelled `migrations.AbstarctMigrationStep`; 0004 correctly uses `AbstractMigrationStep`. Both exist in RA 16, so this is a deliberate improvement, not an inconsistency to fix.

<details>
<summary>Coverage: 8 axes clean · 9 checks run · mode shipped</summary>

**Clean axes:** security — no new tenancy or permission path; project names are already free-form user input through the API, so an apostrophe in an auto-created name adds no injection surface; the audit log at exordos_core/user_api/iam/api/controllers.py:630 already logged usernames; compatibility/upgrade — data-only migration, correctly chained onto 0003, downgrade present; Project carries TargetResourceMixin but nothing in the tree converts IAM models into UA resources (no to_ua_resource call outside config/secret/elements/compute), so no ua_target_resources nudge is owed; diff scope — all four hunks trace back to the stated intent, no adjacent-code tidy-ups; reconciliation/idempotency — no surface in the diff; production readiness — no TODO/WIP/temporary markers, no debug output, no process comments (skill greps over the diff returned nothing); repo hygiene — Apache header in the new migration, English throughout, commit subjects follow <type>(<scope>): <subject>, no generated artefacts committed, no user-API surface change so full_spec.yaml stays current; commit claims — verified: nothing resolves a project by the name 'default' (Project.get_default walks role bindings), migrations.AbstractMigrationStep exists in RA 16, constants.OWNER_ROLE_UUID is the same role Project.add_owner binds; tests — the provisioning behaviour is covered in both the unit and functional suites

**Checks run:**
- git diff --stat / git log / git diff over c24b414...ff53046 — 4 files, 71 insertions, 3 deletions
- grep for project lookups by the literal name 'default' across the repo — only the migration itself and an unrelated token-scope comment; the PR's claim holds
- grep for to_ua_resource / TargetResourceSQLStorableMixin in exordos_core — no IAM model is scheduled into ua_target_resources
- skill axis-9 greps over the diff (TODO/FIXME/WIP/temporary, was/previously/renamed-from comments) — no hits
- skill axis-6 greps over the diff (target-resource surface, migration updated_at nudge) — no target-resource surface
- read restalchemy/storage/sql/migrations.py — AbstractMigrationStep exists at line 52; filename hash and depends convention match
- read restalchemy/api/middlewares/contexts.py — the request runs inside one committed/rolled-back session
- read restalchemy/dm/types.py UTCDateTimeZ — to_simple_type writes a naive UTC string
- extracted DDL for iam_projects / iam_users / iam_organizations / iam_binding_roles from migrations/0000-squashed-current-7f2e4a.py — no unique constraint on project name; name is varchar(128)

</details>

<sub>38 turns · 426s</sub>

exordy (commented):
### exo-review: 1 high: A username longer than 118 characters makes the personal project name overflow iam_proj… · since ff53046: 2 fixed, 1 still open, 0 new

Intent: Name the auto-provisioned personal-workspace project "<username>'s project" instead of the literal "default".

- [thread](https://github.com/exordos/exordos_core/pull/627#discussion_r3945129536) `exordos_core/user_api/iam/dm/models.py:468` **[high]** A username longer than 118 characters makes the personal project name overflow iam_projects.name varchar(128), which rolls back the whole email confirmation.

#### Verify before merge

- The overflow was derived from the column definition and the DM types, not executed. Settle it with a repro against a live DB: INSERT INTO iam_projects (uuid, name, description, organization) with a 129-character name, or POST a registration with a 119-character username and confirm the email on a stand.
- Effect of the planned out-of-band rename on the universal-agent target state: Project mixes in TargetResourceMixin and does not override get_resource_target_fields(), so name is part of the hashed target value, and a hand-written UPDATE iam_projects SET name = ... bypasses the model save path. Rename one project by SQL on a stand and check that its ua_target_resources row's hash and updated_at move.

Note: The branch keeps both ff53046 (adds migration 0004) and 57d89e0 (deletes it). The net diff is correct, but the history ships the experiment — squashing the pair before merge leaves a clean branch.

Note: With the migration dropped, an installation ends up with mixed naming: projects created before the merge stay named 'default', new ones get the owner's name. That is the stated plan, but nothing in the tree performs the rename any more, so the out-of-band step has to actually happen.

<details>
<summary>Coverage: 8 axes clean · 7 checks run · mode shipped</summary>

**Clean axes:** security — no query filter, permission, log line or external parser in the diff; the project name is a stored value on a parameterised insert, and the organization name already carried the same user-controlled string; diff scope — all three hunks trace to the intent, nothing adjacent touched; the revert removed the migration cleanly with no orphan import and no dangling depends (migrations/ ends at 0003); commit claims — 'nothing looks a project up by the name default' verified: the only remaining 'default' in the IAM package is models.py:1698, the project:default scope keyword, which routes to Project.get_default() (earliest role binding); docs/iam/permissions_overview.md:231 documents that keyword and stays correct; compatibility/upgrade — the diff has no migration, manifest, agent config or target-resource shape change and leaves existing rows untouched, so nothing goes stale; reconciliation/idempotency — no surface in the diff; tests — the unit assertion runs against a concrete user.name = 'testuser' and the functional test reads the name back through Project.get_default(); both match the new behaviour; production readiness — no TODO/FIXME, debug prints, process comments or WIP scaffolding; Apache headers intact; repo hygiene — English throughout, commit subjects in <type>(<scope>): <imperative> form, no generated artefacts committed, no dependency changes

**Checks run:**
- git diff --stat / git log / git diff c24b414...HEAD — 3 files, 3 insertions, 3 deletions
- git diff ff53046..HEAD — migrations/0004-rename-default-projects-9af1ca.py deleted in full
- ls migrations/ — sequence ends at 0003, no reference to the deleted 0004
- grep for iam_projects DDL in migrations/0000-squashed-current-7f2e4a.py — name is character varying(128)
- grep for Username( across exordos_core — a single construction, max_length=128
- grep for 'default' as a project name across the repo — only the project:default scope keyword at models.py:1698
- Not run, per the review instructions: tox, unit/functional suites, ruff, mypy (CI covers them)

</details>

<sub>14 turns · 152s</sub>

akremenetsky (approved):

gmelikov (approved):

PR #605, opened 2026-08-25
Add node details action with hypervisor placement

slashburygin:
## Summary

The node→hypervisor placement lives in the `machines` table and was not reachable from the API: the `Node`
resource has no pool field and `machines` has no route. This adds a read-only node action that answers it.

`GET /v1/compute/nodes/<uuid>/actions/details` returns `{"hypervisor": {...}}` with the machine pool the
node was placed on, or `{"hypervisor": null}` while the node has not been scheduled to a machine yet.

## Changes

- `routes.py`: `NodeDetailsActionRoute`, wired as `details` on `NodeRoute` (a GET action, no `/invoke`).
- `controllers.py`: `NodesController.details` finds the node's machine, loads its pool and dumps the whole
  hypervisor record. No extra `_enforce` — the action's resource is fetched through the controller's `get`,
  so the existing node `read` policy already gates it and no new IAM permission is needed.
- Functional test for both the unscheduled and the placed case.
- Troubleshooting guide (en + ru): documented next to the existing node/hypervisor checks.

Note for review: `/v1/compute/hypervisors/` is admin-scoped, while this action is gated only by the node
`read` policy — so the embedded record hands a project user the pool's `driver_spec` (`connection_uri`,
`machine_prefix`). Say the word if the dump should be narrowed to a field whitelist or get its own policy.

## Verification

- `pytest exordos_core/tests/unit -n 10` — 261 passed
- `DATABASE_URI=... pytest exordos_core/tests/functional -n 10` — 534 passed, 3 skipped
- `ruff check exordos_core`, `ruff format --check` — clean
- `make mdlint` — 0 errors
- `ec-manifest-schema` — no diff (a GET action is not a create path, so `full_spec.yaml` is unchanged)

https://claude.ai/code/session_01GqRBZ3ueJbBvhY9iX3wo7A

## Summary by Sourcery

Expose node hypervisor placement through the compute API.

New Features:
- Add a read-only node details action that reports the hypervisor pool hosting the node, or null when it is unscheduled.

Documentation:
- Document the node placement details action in English and Russian troubleshooting guides.

Tests:
- Add functional coverage for both unscheduled nodes and nodes placed on a machine.

Chores:
- Register the node details permission and owner binding in the core manifest.

PR #637, opened 2026-09-11
Add opaque secret resource

slashburygin:
## Summary

Adds `$core.secret.secrets`: an opaque secret whose value the platform does not
interpret. An element can consume an API token, license key or connection
string by reference instead of carrying it in its manifest.

- The value is write only in the user API. It can be set and rotated, but no
  response returns it (read, list, create or update).
- A manifest can ship a `default_value` that stands in until an operator sets a
  real value. Clearing the value falls back to the default.
- A secret with neither holds in `NEW`, and an element that renders `:value`
  from it waits instead of deploying with an empty secret.

The branch also fixes a pre-existing convergence bug that the new resource
exposed. A data plane backend client's `update()` refreshed only the field it
regenerated, so renaming a password or certificate left the stored `meta` hash
permanently apart from the target, and the resource sat in `NEW` forever.

## Changes

**`feat(secret): add opaque secret resource`**

- `Secret` control plane model and `secret_secrets` table; data plane model,
  `storage_secrets` table and backend client
- `/v1/secret/secrets/` with `value` hidden, and an update that returns the
  secret to `NEW` so it is redelivered
- `SecretBuilder` and `OpaqueSecretCapabilityDriver`, registered with the
  scheduler and the core agent model map, both in the config templates and in
  the `ec-bootstrap` upgrade path (upgraded stands restore configs from the
  persisted copy)
- Quota, telemetry and `TABLES_TO_MODELS` entries, and `secret.secret.*` IAM
  permissions
- Migration 0003 guarded with `to_regclass`, so tables added to
  `TABLES_TO_MODELS` after it still migrate on a fresh database
- Docs page, EM resource reference and a manifest example

**`fix(secret): refresh the whole meta on a data plane update`**

- The password client rebuilds `meta` from the target resource, so a rename
  converges without rotating the password
- The certificate client refreshes `meta` on its no-renewal path
- `test_rename_password`, plus `TestCertBackendClient` covering the real
  `CertBotBackendClient` (the builder tests use a fake that hid the bug)

**`feat(secret): add a default value to opaque secrets`**

- `value` becomes optional and `default_value` is added. The model view reports
  the effective value, so the data plane and `:value` both see the default
- The builder readiness hook holds a secret with no effective value in `NEW`
- A resource transformer drops a null `value` from the reported state, so a
  consuming element waits rather than rendering null; the section is added on
  upgraded stands too
- Troubleshooting entry for an element stuck on a valueless secret (EN/RU)

**`fix(secret): drop the project filter from the secret capability`**

- A manifest secret belongs to the manifest's project, not the service project.
  Scoping `em_core_secret_secrets` to one project hid those rows from the core
  agent, which retried the create and conflicted on the primary key forever
- The filter is removed from the template, `ec-bootstrap` strips the line older
  images wrote, and the model-mapping check anchors on the mapping itself
- `project_id` is now required in a manifest secret (it used to be injected
  from the filter)

**`feat(iam): grant the owner role opaque secret permissions`**

- Bind `secret.secret.read` and `secret.secret.update` to the owner role. A
  secret is declared by a manifest and goes away with it, so an owner reads the
  record and sets its value, but neither creates nor deletes one

**`fix(secret): close the review findings on opaque secrets`**

- Create and update responses no longer include `value` or `default_value`.
  Previously `PUT {"name": ...}` returned the stored secret to anyone with
  update permission
- Clearing the value of a delivered secret with no default now withdraws it:
  the builder deletes the target resource, the agent drops the stored value and
  the secret waits in `NEW`. Before, the update was held forever with the old
  plaintext still on the node
- `ec-bootstrap` inserts the secret model mapping after `[models]` and appends
  the value transformer when the persisted config has no password lines to
  anchor on
- A password is regenerated when its method changes, not only its length
- `secret_secrets_project_id_idx`, matching the other secret tables
- Manifest example fixed to a valid `kind: text` body with an inline
  `f"...{...:value}"` reference; `:value` no longer documented as read from the
  data plane

## Verification

Run on `0fc7293`:

- Functional suite: 565 passed, 3 skipped (`pytest -n 10`, local PostgreSQL)
- Unit suite: 281 passed
- `ruff check .`: clean
- `make mdlint`: clean
- The three regression tests added in the last commit (update response leak,
  value withdrawal, password method change) fail with the source fixes reverted
  and pass with them
- The rename tests from the meta fix were confirmed red with that fix reverted
- `tox -e mypy` was not run: the env does not build locally and it is not part
  of CI

## Follow-ups not in this PR

- **Permissions and owner bindings on upgraded stands.** The new
  `secret.secret.*` permissions and owner bindings live only in the core
  manifest. `ec-bootstrap` skips reapplying it once the core element is
  installed, so existing stands get them only through a core element upgrade.
  This matches how #605 and the repo artifacts permissions were added. Seeding
  them from a migration could collide with the manifest later creating the same
  UUIDs, so it is left as a separate decision
- `docs/em/resources.ru.md` still says `:value` reads the value delivered to the
  data plane; the English page is fixed
- Migration 0003 is edited in place. It is inert on stands that already applied
  it and required on fresh databases, but worth a reviewer's eye
- A non-`plain` constructor would break hash convergence, since `meta["value"]`
  holds the built value while the target field holds the plaintext. Only
  `plain` exists today, and passwords share the same shape

https://claude.ai/code/session_01HK91UscQ3JiXNnEXgTw1ge


docs/em/resources.md
```diff
@@ -862,6 +863,43 @@ resources:
...
+      constructor:
+        kind: plain
+      value: "glsa_XXXXXXXXXXXXXXXX"
+```
+
+### Notes
+
+- The value is write only in the user API: it can be set and replaced, but it is never returned by a
+  read. Reference it from a manifest with the `:value` link parameter, which reads the value delivered
+  to the data plane, or reference the secret itself with `:uuid`.
```
exordy: **[low] `:value` is documented as reading the value delivered to the data plane; it resolves from the control-plane row.**

SecretBuilder.actualize_outdated_instance syncs status only — the value never returns from the DP. A reader concludes the reference is unresolvable until the secret reaches ACTIVE, and that a non-plain constructor would yield the built value.

</details>

```suggestion
  read. Reference it from a manifest with the `:value` link parameter, or reference the secret itself
  with `:uuid`.
```

exordy (commented):
### exo-review: 2 low, nothing blocking

Intent: Add the opaque `$core.secret.secrets` resource end to end, and fix DP `update()` leaving stale `meta`.

#### Outside the diff

- `exordos_core/cmd/bootstrap.py:447` **[low]** The function's docstring still lists only BorderAgentCapabilityDriver as the caps_drivers line it inserts. **Fix:** Name both drivers in step 2. <details><summary>Why</summary>

A numbered list that enumerates what the function reapplies now omits one of its four steps, so the next reader of an upgrade path with no unit coverage has to diff the body against the list.

</details> 

#### Verify before merge

- Fresh-DB chain: apply 0000→0004 on an empty database, then \d secret_secrets to confirm the tags column and GIN index exist.
- ec-bootstrap against a persisted core_agent.conf that does carry the em_core_secret_passwords anchor — no unit coverage either way.
- Manifest rendering of $core.secret.secrets.$x:value on a stand; nothing in-tree exercises the link parameter for the new kind.

Note: gcl_sdk agent.py logs str(r.value) when create() raises — a failed secret create puts the plaintext in the agent log.

Note: Sourcery's three curl-auth-header hits on docs/secret/secrets.md are the literal MY_TOKEN placeholder; nothing to redact.

<details>
<summary>Coverage: 9 axes clean · 8 checks run · mode shipped</summary>

**Clean axes:** security — value HIDDEN on GET/FILTER, EM controllers hide target/actual_resource; tenancy — project_id is a target field, policy controller mirrors passwords; scope — every hunk traces to the new resource or to the meta fix; compat — 0004 additive; password/cert target shapes unchanged, no nudge owed; reconciliation — update() only runs for rows list() returned, so it converges; contracts — Secret→AbstractSecret rename: every in-tree reference updated; tests — rename tests for secret, password and the real CertBotBackendClient; prod-readiness — no TODO/WIP/debug/process-comment leftovers in the diff; hygiene — Apache headers, English, commit format, full_spec entry present

**Checks run:**
- read full patch (2004 lines) + stat + log from the bundle
- grep patch for TODO/WIP/for-now/print/sleep — clean
- grep patch for was-X-now-Y comments — clean
- grep tree for models.Secret/driver_dm.Secret refs — all updated
- read gcl_sdk direct.py + agent.py: update() unreachable without a DP row
- checked quota/telemetry/TABLES_TO_MODELS/entry-point/scheduler registration
- checked migration numbering, enum_secret_status, 0003 to_regclass guard
- no tests/linters run (CI owns them, per instructions)

</details>

<sub>35 turns · 589s</sub>

exordos_core/secret/builders/service.py
```diff
@@ -25,6 +25,22 @@
...
+):
+    @classmethod
+    def get_resource_kind(cls) -> str:
+        return sc.SECRET_KIND
+
+    def is_ready_to_actualize(self) -> bool:
+        # A secret with neither a value nor a default has nothing to
+        # deliver. Hold it in NEW instead of pushing an empty secret to
+        # the data plane: an element consuming it waits for the value.
+        return self.effective_value is not None
```
exordy, line 41: **[medium] Clearing the value of an already delivered secret that has no default withdraws nothing and never converges.**

**Fix:** Withdraw rather than hold: drop the target resource when the effective value disappears, or refuse to clear a value with no default behind it.

ReadinessMixin routes is_ready_to_update here, and a false gate skips _actualize_updated_instance entirely, so the target resource and the node's storage_secrets row keep the previous value while the controller has already put the row back to NEW. Nothing revisits it.

</details>

exordy (commented):
### exo-review: 1 medium, 2 low, nothing blocking · since c54c307: 0 fixed, 2 still open, 1 new

- [thread](https://github.com/exordos/exordos_core/pull/637#discussion_r3986163998) `docs/em/resources.md:899` **[low]** `:value` is still documented as reading the value delivered to the data plane; it resolves from the control-plane row.
- `exordos_core/cmd/bootstrap.py:447` **[low]** The docstring's numbered list still omits the OpaqueSecretCapabilityDriver insertion the body now performs.

#### Verify before merge

- The core agent config rewrite against a real persisted file: run _ensure_core_agent_config_current over a copy of a stand's /etc/exordos_core/core_agent.conf.
- 0003 + 0004 on an upgraded database: ec-migrate apply on a dump of a stand that already has 0003.

Note: The new transformer insertion depends on the password section anchor, same as the model/filter lines the bots flagged.

<details>
<summary>Coverage: 9 axes clean · 6 checks run · mode shipped</summary>

**Clean axes:** security — value and default_value both HIDDEN on GET/FILTER, tests assert it; tenancy — project_id target field, controller mirrors passwords; scope — every hunk traces to the resource, the default or the meta fix; compat — 0004 additive; 0003 guard inert on applied stands; no target shape change owing a nudge; reconciliation — EM update writes only manifest-declared keys, so an operator value is not reverted; contracts — default reaches the DP through dump_to_simple_view; target fields unchanged for other secrets; tests — default fallback, hold, rotate, rename and delete covered functionally; prod-readiness — no TODO/WIP/debug leftovers; hygiene — Apache headers, English, commit format, full_spec regenerated

**Checks run:**
- read the whole pr-637 bundle in slices — no new hunk outside the intent
- gcl_sdk builder.py: can_update gate skips the whole update — confirms finding 1
- gcl_sdk db.py _update: only keys present in the target are written
- gcl_sdk direct.py transformer: hash uses the target-shaped keys only
- grep docs/em/resources.md:898 — previous finding 2 wording unchanged
- no tests, linters or type checkers run (CI owns them, per instructions)

</details>

<sub>20 turns · 356s</sub>

exordos/manifests/core.yaml.j2
```diff
@@ -748,6 +764,22 @@ resources:
     binding_owner_repo_artifact_read:
       role: "$core.iam.roles.$owner:uuid"
       permission: "$core.iam.permissions.$repo_artifact_read:uuid"
+    binding_owner_secret_secret_read:
```
exordy, line 767: **[medium] The new owner-role secret bindings are manifest-only and never reach a stand whose core RepoElement is already ACTIVE.**

**Fix:** Seed the new permissions and bindings from the upgrade path itself — same fix as the already-open permission-seeding gap, now covering bindings too.

ec-bootstrap's manifest reapply for the core repo returns early once its RepoElement is ACTIVE/IN_PROGRESS/ERROR (bootstrap.py:825-832) — the same short-circuit already flagged for secret.secret.* permissions now also blocks these bindings.

</details>

exordy (commented):
### exo-review: 2 medium, 2 low, nothing blocking · since 3ce6208: 0 fixed, 3 still open, 1 new

- [thread](https://github.com/exordos/exordos_core/pull/637#discussion_r3988793697) `exordos_core/secret/builders/service.py:41` **[medium]** Clearing the value of an already-delivered secret with no default never converges.
- [thread](https://github.com/exordos/exordos_core/pull/637#discussion_r3986163998) `docs/em/resources.md:899` **[low]** `:value` is still documented as reading the value delivered to the data plane; it resolves from the control-plane row.
- `exordos_core/cmd/bootstrap.py:447` **[low]** _ensure_ua_config_current's docstring still lists 3 steps while the body also inserts OpaqueSecretCapabilityDriver.

#### Verify before merge

- Whether an ec-bootstrap run on a DB with an already-ACTIVE core RepoElement actually leaves iam_binding_permissions short the 4 new rows

Note: New owner bindings extend the same gap chatgpt-codex-connector flagged for secret.secret.* permissions (core.yaml.j2:488) — one upgrade-path fix covers both

<details>

**Clean axes:** scope — both hunks trace to intent: drop stale filter, bind owner role; security — bindings follow the $compute_core owner-scoping convention; compat — filter removal and docstring update are additive/idempotent, tested; tests — test_drops_secret_project_filter fails if reverted; idempotency covered; hygiene — commit format/English clean, no full_spec impact (data, not schema); prod-readiness — no TODO/WIP/debug leftovers in this range

**Checks run:**
- grep bootstrap.py upgrade short-circuit on RepoElement status
- read secret/builders/service.py:37-41 — finding 1 unchanged
- grep docs/em/resources.md:899 — finding 2 unchanged
- read bootstrap.py:434-451 — finding 3 still 3 steps
- diffed new owner bindings vs dns_domain/network_lb siblings

</details>

<sub>32 turns · 273s</sub>

exordy (commented):
### exo-review: 2 medium, 2 low, nothing blocking · since d028fee: 0 fixed, 4 still open, 0 new

- [thread](https://github.com/exordos/exordos_core/pull/637#discussion_r3995332332) `exordos/manifests/core.yaml.j2:767` **[medium]** Owner-role secret bindings are manifest-only and never reach a stand whose core RepoElement is already ACTIVE.
- [thread](https://github.com/exordos/exordos_core/pull/637#discussion_r3988793697) `exordos_core/secret/builders/service.py:41` **[medium]** Clearing the value of an already-delivered secret with no default never converges.
- [thread](https://github.com/exordos/exordos_core/pull/637#discussion_r3986163998) `docs/em/resources.md:899` **[low]** `:value` is still documented as reading the value delivered to the data plane; it resolves from the control-plane row.
- `exordos_core/cmd/bootstrap.py:447` **[low]** _ensure_ua_config_current's docstring still lists 3 steps while the body also inserts OpaqueSecretCapabilityDriver.

<details>
<summary>Coverage: 5 axes clean · 4 checks run · mode shipped</summary>

**Clean axes:** scope — 4-line removal, matches commit message exactly; security — project_id unused by enforcement, verified in migrations/0000 view SQL; consistency — matches existing owner_iam_*/owner_repo_* bindings; compat — removing an always-ignored field is a no-op; hygiene — commit format/English clean, single-purpose refactor

**Checks run:**
- grep core.yaml.j2 for binding_owner_iam_*/repo_* to confirm convention match
- grep+read iam_permissions_fast_view SQL in migrations/0000 to confirm project_id unused
- read bootstrap.py:825-832 short-circuit — unchanged, finding 1 still open
- read secret/builders/service.py:41, docs/em/resources.md:899, bootstrap.py:447 — all unchanged

</details>

<sub>15 turns · 79s</sub>

exordy (commented):
### exo-review: 3 medium, 2 low, nothing blocking · since f3203f3: 0 fixed, 4 still open, 1 new

#### Outside the diff

- `exordos_core/elements/dm/models.py:1895` **[medium]** A `secret` import with no `project_id` searches every project's secrets by name, with no ownership check. **Fix:** Require `project_id` on a secret import, as `$core.secret.secrets` already requires it on a declared secret. <details><summary>Why</summary>

`_get_secret` filters `Secret.objects.get_all()` by name only when `project_id` is absent; the DM layer applies no tenant scoping, so a unique name anywhere resolves and its value renders into the target state.

</details> 

- [thread](https://github.com/exordos/exordos_core/pull/637#discussion_r3995332332) `exordos/manifests/core.yaml.j2:767` **[medium]** Owner-role secret bindings are manifest-only and never reach a stand whose core RepoElement is already ACTIVE.
- [thread](https://github.com/exordos/exordos_core/pull/637#discussion_r3988793697) `exordos_core/secret/builders/service.py:37` **[medium]** Clearing the value of an already-delivered secret with no default never converges.
- [thread](https://github.com/exordos/exordos_core/pull/637#discussion_r3986163998) `docs/em/resources.md:899` **[low]** `:value` is still documented as reading the value delivered to the data plane; it resolves from the control-plane row.
- `exordos_core/cmd/bootstrap.py:435` **[low]** `_ensure_ua_config_current`'s docstring still lists 3 steps while the body also inserts `OpaqueSecretCapabilityDriver`.

#### Verify before merge

- test_full_spec_is_current actually passing — run tox -e py312 -- -k test_full_spec_is_current (CI covers it)
- Whether any downstream manifest outside this repo grants a non-admin role em.manifest.*

Note: Cross-project secret-import lookup is deliberately designed and tested, not an oversight — flagged per the repo's ownership-check convention.

Note: chatgpt-codex-connector's P1 on controllers.py:53 (value leaks on metadata-only update) is unresolved but already raised; not repeated.

<details>
<summary>Coverage: 7 axes clean · 5 checks run · mode shipped</summary>

**Clean axes:** scope — every hunk in the new commit traces to importing an API-created secret; schema — oneOf import schema rejects a half-edited copy, tested both ways; tests — unit+functional cover value/default/missing/ambiguous/cross-project cases; docs — en/ru parity kept across manifest.md, resources.md, exports.md, secrets.md; reconciliation — KeyError reuses the existing hold-in-place pattern, nothing new; hygiene — commit format/English clean, Apache headers on new test files; migration — em_imports schema change has a down migration; 0004 still unshipped

**Checks run:**
- grep core.yaml.j2 binding_owner_secret_secret_* — present, unbound at bootstrap short-circuit
- grep em.manifest bindings across manifests — none, install stays admin-only today
- read ReadinessMixin in gcl_sdk — confirms is_ready_to_update routes to is_ready_to_actualize
- read base_spec.yaml oneOf schema — resource_import kind excludes 'secret', secret_import rejects extras
- git log f3203f33d1..HEAD on 4 previously-flagged files — none touched, findings carried forward

</details>

<sub>46 turns · 387s</sub>

exordy (commented):
### exo-review: 2 medium, 2 low, nothing blocking · since fb07cdc: 1 fixed, 4 still open, 0 new

- [thread](https://github.com/exordos/exordos_core/pull/637#discussion_r3995332332) `exordos/manifests/core.yaml.j2:767` **[medium]** Owner-role secret bindings are manifest-only and never reach a stand whose core RepoElement is already ACTIVE.
- [thread](https://github.com/exordos/exordos_core/pull/637#discussion_r3988793697) `exordos_core/secret/builders/service.py:37` **[medium]** Clearing the value of an already-delivered secret with no default never converges.
- [thread](https://github.com/exordos/exordos_core/pull/637#discussion_r3986163998) `docs/em/resources.md:899` **[low]** :value is documented as reading the value delivered to the data plane; it actually resolves from the control-plane row.
- `exordos_core/cmd/bootstrap.py:435` **[low]** _ensure_ua_config_current's docstring lists 3 reapply steps while the body also inserts OpaqueSecretCapabilityDriver.

#### Verify before merge

- Whether enforcement (iam_permissions_fast_view) truly ignores project_id end-to-end, per the commit's rationale — read the permission-check query path in user_api/iam

Note: PR body describes an em-import-secrets commit (16 tests, docs) that isn't in this branch's history — replaced by a single e7f68ad commit.

Note: The bundled patch reflects the stale fb07cdc101 base; real range 62c3be5..HEAD is 6 lines. Flag the description/branch mismatch before merging.

<details>
<summary>Coverage: 6 axes clean · 3 checks run · mode shipped</summary>

**Clean axes:** scope — 6-line diff binds exactly read+update, matching the commit message's stated design; commit claims — verified: no create/delete bound, matches 'neither creates nor deletes one'; hygiene — commit format compliant, imperative subject, English; compat — no migration/API/schema/target-resource shape touched by this commit; reconciliation — no reconciliation code touched; tests — declarative manifest addition, consistent with untested sibling bindings in the same file

**Checks run:**
- git merge-base fb07cdc101 HEAD → 62c3be5; fb07cdc101 not an ancestor of HEAD
- git diff 62c3be5..HEAD --stat → only core.yaml.j2, 6 lines
- grep secret_secret_* bindings in core.yaml.j2 → read+update only, confirmed

</details>

<sub>17 turns · 204s</sub>

docs/secret/secrets.md
```diff
@@ -0,0 +1,98 @@
...
+Secrets are a part of the Secret Manager service. A secret holds an opaque value the platform does not
+interpret: an API token, a license key, a connection string, anything an element needs but should not
+carry in its manifest.
+
+Unlike passwords, a secret has no generation method — the value is always supplied by the user.
+
+```bash
+curl --location 'http://10.20.0.2:11010/v1/secret/secrets/' \
+--header 'Content-Type: application/json' \
+--header 'Authorization: Bearer MY_TOKEN' \
```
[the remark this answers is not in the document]
slashburygin: False positive: `MY_TOKEN` is a documentation placeholder, not a real credential. It is the same convention the other Secret Manager docs pages use.

docs/secret/secrets.md
```diff
@@ -0,0 +1,98 @@
...
+  format.
+- **value** - the opaque value. Required, up to 10240 characters.
+
+## The value is write only
+
+The value can be set on create and replaced on update, but it is never read back through the API:
+
+```bash
+curl --location 'http://10.20.0.2:11010/v1/secret/secrets/<uuid>' \
+--header 'Authorization: Bearer MY_TOKEN'
```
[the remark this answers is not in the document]
slashburygin: False positive: `MY_TOKEN` is a documentation placeholder, not a real credential. It is the same convention the other Secret Manager docs pages use.

docs/secret/secrets.md
```diff
@@ -0,0 +1,98 @@
...
+```
+
+There is no `value` field in the response, and no way to ask for one. The only response the value ever
+appears in is the reply to the create or update request that supplied it. Rotating a secret is a
+regular update:
+
+```bash
+curl --request PUT --location 'http://10.20.0.2:11010/v1/secret/secrets/<uuid>' \
+--header 'Content-Type: application/json' \
+--header 'Authorization: Bearer MY_TOKEN' \
```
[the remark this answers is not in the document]
slashburygin: False positive: `MY_TOKEN` is a documentation placeholder, not a real credential. It is the same convention the other Secret Manager docs pages use.

docs/secret/secrets.md
```diff
@@ -0,0 +1,98 @@
+---
+icon: lucide/lock
+---
+# Secrets
+
+Secrets are a part of the Secret Manager service. A secret holds an opaque value the platform does not
+interpret: an API token, a license key, a connection string, anything an element needs but should not
+carry in its manifest.
```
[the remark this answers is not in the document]
slashburygin: Not applicable: the README has no per-service feature list (passwords, certificates and RSA/SSH keys are not listed there either), so there is nothing to keep in sync. The resource is documented in `docs/secret/secrets.md`, `docs/em/resources.md` and mkdocs nav.

exordos_core/elements/dm/models.py
```diff
@@ -1043,6 +1072,127 @@ def link(self) -> str:
         return f"{self.element.link}.imports.${self.name}"
 
 
+class ImportedSecret:
```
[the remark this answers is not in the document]
slashburygin: Obsolete: secret imports were dropped from this PR. `ImportEnum` only has `RESOURCE` again and `elements/dm/models.py` no longer contains the imported-secret class.

exordos_core/elements/dm/models.py
```diff
@@ -1043,6 +1072,127 @@ def link(self) -> str:
...
+        filters = {"name": ra_filters.EQ(self._secret_name)}
+        if self._project_id is not None:
+            filters["project_id"] = ra_filters.EQ(self._project_id)
+
+        secrets = secret_models.Secret.objects.get_all(filters=filters)
+        if len(secrets) > 1:
+            # Only a name-only import can match more than one secret. The
+            # manifest has to say which project it means; picking one here
+            # would hand the element another project's secret.
+            raise exceptions.ValidateException(
```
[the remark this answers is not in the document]
slashburygin: Obsolete: secret imports were dropped from this PR, so the name-only secret lookup and its ambiguity exception are gone.

exordy (commented):
### exo-review: 1 medium, 1 low, nothing blocking · since e7f68ad: 2 fixed, 2 still open, 0 new

- [thread](https://github.com/exordos/exordos_core/pull/637#discussion_r3995332332) `exordos/manifests/core.yaml.j2:767` **[medium]** Owner-role secret permission bindings still never reach a stand whose core RepoElement is already ACTIVE.
- `exordos_core/cmd/bootstrap.py:434` **[low]** _ensure_ua_config_current's docstring still lists 3 reapply steps while the body inserts a 4th (OpaqueSecretCapabilityDriver).

#### Verify before merge

- Full functional/unit suite result for this commit — trust CI per instructions
- resource.delete() cascade under pytest-xdist parallel workers — covered only in isolation by the new test

<details>
<summary>Coverage: 8 axes clean · 3 checks run · mode shipped</summary>

**Clean axes:** codex P1 leak — _hide_values blanks value/default_value on create+update, tested; codex P1 upgrade anchors — [models]/transformer fallback verified vs anchorless fixture; codex P2 index — secret_secrets_project_id_idx matches sibling convention; codex P2 password method — meta method comparison added, tested; codex P2 doc example — valid kind:text/content with f"..." now used; reconciliation — delete() drops instance from updated-instances join next pass, no race; hygiene — commit format compliant, Apache headers unchanged, English; scope — every hunk maps to a listed prior finding

**Checks run:**
- Read gcl_sdk builder.py/models.py to verify can_update_instance_resource + get_updated_entities semantics
- Grepped core.yaml.j2 for secret.secret bindings — unchanged, finding 1 still open
- Verified Config.body kind:text/content shape and f"..." convention against docs/core-developer-guide

</details>

<sub>35 turns · 260s</sub>

docs/em/resources.md
```diff
@@ -862,6 +863,48 @@ resources:
...
+
+```yaml
+resources:
+  $core.secret.secrets:
+    grafana_api_token:
+      name: "grafana-api-token"
+      project_id: "12345678-c625-4fee-81d5-f691897b8142"
+      constructor:
+        kind: plain
+      default_value: "glsa_XXXXXXXXXXXXXXXX"
```
phantomii, line 892: Could we rename `constructor` to `content` and move `value` and `default_value` inside it?

```yaml
content:
  kind: plain
  value: "glsa_OPERATOR_TOKEN"
  default_value: "glsa_DEFAULT_TOKEN"
```

Then `content.kind` would discriminate the secret content type (plain text, certificate, etc.), and the corresponding values would be validated by that variant.

exordos/manifests/core.yaml.j2
phantomii: Confirmed: this risk is still present on the current head. Existing installations skip the core manifest reapply, so the new permissions and owner bindings are never seeded.

phantomii (changes requested):
Please address the existing-installation upgrade path before merging: the bootstrap skips reapplying the core manifest for an already active core element, while the migration does not seed the new Secret permissions and owner bindings.

docs/em/resources.md
slashburygin: Two things worth settling before I make this change.

`constructor` isn't new here — it's declared on `AbstractSecret` (`exordos_core/secret/dm/models.py`) and is already part of the released API for passwords, certificates, RSA keys and SSH keys. So the rename is either:

- opaque secrets only, which leaves `$core.secret.secrets` on `content` while the other four stay on `constructor`; or
- all secret types, which is a breaking change to the released manifest and REST API, and needs the `constructor` column renamed across `secret_passwords`, `secret_certificates`, `secret_rsa_keys` and `secret_ssh_keys`, plus the corresponding doc and test updates.

Separately, moving `value` and `default_value` inside `content` breaks the write-only guarantee. Both are currently top-level properties carrying `Permissions.HIDDEN` for GET and FILTER, and restalchemy's `FieldsPermissions.meets_field_permission` matches a flat `model_field_name` — it has no nested-path support. Once those values live inside a single `content` jsonb, nothing hides them and a GET returns the secret. The same flat assumption shows up in the core agent config, where the `em_core_secret_secrets` transformer is declared as `attributes = value`.

Keeping them top-level gets the `content.kind` discriminator without that regression:

```yaml
content:
  kind: plain
value: "glsa_OPERATOR_TOKEN"
default_value: "glsa_DEFAULT_TOKEN"
```

Full nesting as written would need hand-written masking in the controller to replace what the field permissions do today. Which would you prefer?

migrations/0004-add-opaque-secrets-362f6d21.py
```diff
@@ -0,0 +1,178 @@
...
+    ON CONFLICT (uuid) DO NOTHING
+    """
+    for permission_uuid, name, description in SECRET_PERMISSIONS
+]
+
+# A stand that did reapply the manifest already has the binding under a
+# UUID of its own, so match on the pair instead of the UUID.
+UPGRADE += [
+    f"""
+    INSERT INTO iam_binding_permissions (uuid, role, permission)
```
exordy: **[medium] The pair guard dedupes only one ordering: on a fresh install the core manifest adds a second binding row for each pair.**

**Fix:** Give the two bindings in core.yaml.j2 explicit uuids equal to OWNER_BINDINGS, as the permissions already do.

Manifest bindings declare no uuid, so the engine assigns a fresh uuid4 (common/utils.py:89). Migrations run before the manifest, so NOT EXISTS matches nothing and the agent inserts its own row next to the seeded one.

</details>

exordy (commented):
### exo-review: 1 medium, 1 low, nothing blocking · since 0fc7293: 1 fixed, 1 still open, 1 new

- `exordos_core/cmd/bootstrap.py:434` **[low]** _ensure_ua_config_current's docstring still lists 3 reapply steps while the body inserts a 4th.

#### In reply to your comments

- [`exordos/manifests/core.yaml.j2:767`](https://github.com/exordos/exordos_core/pull/637#discussion_r3995332332) @phantomii: Migration 0004 now seeds the four permissions under the manifest's own UUIDs and the two owner bindings, so an upgraded stand gets them without the manifest reapply.

#### Verify before merge

- Fresh install: after bootstrap, SELECT role, permission, count(*) FROM iam_binding_permissions GROUP BY 1,2 HAVING count(*)>1 — confirms the duplicate above.
- Upgraded stand: run the migration, then PUT /v1/secret/secrets/<uuid>/ with an owner token to confirm the seeded grant is effective.

<details>
<summary>Coverage: 7 axes clean · 5 checks run · mode shipped</summary>

**Clean axes:** security — global bindings, same scope the manifest declares; compat — 0004 is new in this branch, edit in place is safe; compat — downgrade deletes bindings before permissions, FK order right; reconciliation — same-uuid create hits ResourceAlreadyExists then driver.get; schema — status/description/tags all defaulted, inserts valid; scope — the one commit maps to the reviewer's upgrade request; hygiene — commit format, English, constants import matches 0001/0003

**Checks run:**
- Read agent.py/direct.py/db.py in gcl_sdk: create→get fallback on conflict
- grep core.yaml.j2: the 4 permission UUIDs match the migration exactly
- grep 0000-squashed: iam_permissions_name_idx is UNIQUE, owner role seeded
- grep get_or_create_uuid_from_dict: uuidless manifest resources get uuid4
- grep 0003: tags is NOT NULL DEFAULT '{}', inserts unaffected

</details>

<sub>42 turns · 409s</sub>

exordos/manifests/core.yaml.j2
slashburygin: Addressed in a6221d9 and ac8b12e — the upgrade path no longer depends on the manifest reapply.

Migration 0004 now seeds the four `secret.secret.*` permissions and the two owner bindings itself:

- The permissions are seeded under the UUIDs the manifest declares. `iam_permissions_name_idx` is unique, so seeding under a derived UUID would collide with the manifest on a fresh installation.
- The bindings are seeded with `project_id` unset, matching what the manifest declares, and the downgrade deletes the bindings before the permissions.

The first cut of this guarded the binding insert on the `(role, permission)` pair, which only covered one ordering: migrations run before the manifest, so on a fresh installation the guard matched nothing and the engine added a second row next to the seeded one, since a manifest binding that declares no uuid gets a fresh uuid4 (`common/utils.py:88`). The two bindings in `core.yaml.j2` now declare explicit uuids equal to the seeded ones, as the permissions and `admin_allow_all` already do, so both orderings converge on a single row: the uuid covers migration-first, the pair guard covers a stand that reapplied the manifest first and holds the binding under a uuid of its own.

Verified by applying the migration to a scratch database: four permissions and two bindings with `role = owner` and `project_id` NULL, re-running the seeding adds no duplicates, a pre-existing manifest-style binding keeps its own uuid, and the downgrade removes everything. The edited manifest validates against both the base and the full manifest schema.

One thing I did not touch: the bootstrap short-circuit affects any manifest-declared IAM resource, not just these. `0000-squashed` seeds the repo, border, compute and DNS bindings with derived uuids while `core.yaml.j2` declares those same bindings without one, so the duplicate-row shape above looks like it applies there too. Out of scope for this PR, but worth a separate look.

exordy (approved):
### exo-review: 1 low, nothing blocking · since a6221d9: 1 fixed, 1 still open, 0 new

- `exordos_core/cmd/bootstrap.py:451` **[low]** _ensure_ua_config_current's docstring still lists 3 reapply steps while the body performs 4.

#### Verify before merge

- Fresh install converges on one row per pair: after ec-bootstrap, SELECT role, permission, count(*) FROM iam_binding_permissions WHERE permission IN (the 4 secret uuids) GROUP BY 1,2.

<details>
<summary>Coverage: 7 axes clean · 4 checks run · mode shipped</summary>

**Clean axes:** security — global bindings, same scope the manifest declares; compat — 0004 new in this branch, edit in place safe; compat — downgrade deletes bindings before permissions, FK order right; reconciliation — same-uuid create hits ResourceAlreadyExists then driver.get; schema — status/description/project_id defaults match the seeded row; scope — both hunks are the pinning, nothing else; hygiene — commit format, English, uuids used nowhere else

**Checks run:**
- grep both uuids repo-wide: migration and manifest agree, no third use
- read common/utils.py:88: a declared uuid is used verbatim
- grep 0000-squashed: (role, permission) is an index, not a unique key
- read bootstrap.py:435-500: docstring list still at 3 steps

</details>

<sub>7 turns · 115s</sub>

phantomii (approved):

exordos/manifests/core.yaml.j2
```diff
@@ -748,6 +764,14 @@ resources:
...
+      uuid: "4f3a8e13-9bea-52f2-b2ce-3acb390a8d99"
```
akremenetsky: Why we would like to specify UUID here and for `binding_owner_secret_secret_update` as well?

exordos_core/cmd/bootstrap.py
```diff
@@ -530,31 +565,78 @@ def _ensure_core_agent_config_current() -> None:
...
-    )
-    new_content = new_content.replace(
-        "em_core_dns_domains_records = "
-        "project_id:12345678-c625-4fee-81d5-f691897b8142\n",
-        "em_core_dns_domains_records = "
-        "project_id:12345678-c625-4fee-81d5-f691897b8142\n"
-        + _CORE_AGENT_IDP_FILTER_LINE,
-        1,
-    )
+    if _CORE_AGENT_SECRET_MODEL_LINE not in new_content:
```
akremenetsky: Let's refactor this part otherwise it will be hard to maintain. I suggest add a funciton something like 
```python
def _ensure_section(cfg_path: pathlib.Path, section: str, items: dict)
```

And explicitly specify what should be in each section.

exordos_core/gservice/service.py
```diff
@@ -195,6 +195,9 @@ def __init__(self, iter_min_period=3, iter_pause=0.1):
         secret_svc = secret_service.SecretServiceBuilder(
```
akremenetsky, line 195: Do we use anything from this builder?

exordos_core/user_api/secret/api/controllers.py
```diff
@@ -29,6 +29,57 @@ class SecretController(controllers.RoutesListController):
...
+        convert_underscore=False,
+        fields_permissions=field_p.FieldsPermissions(
+            default=field_p.Permissions.RW,
+            fields={
+                "status": {ra_c.ALL: field_p.Permissions.RO},
+                # The value is write only: it may be set and rotated but
+                # it is never read back. The default behind it is a
+                # secret value too, so it is hidden the same way. Create
+                # and update responses drop both in `_hide_values`.
+                "value": {
```
akremenetsky: It's not obvious how PaaS services will be working with this approach. They cannot implement full reconciliation.

migrations/0004-add-opaque-secrets-362f6d21.py
```diff
@@ -0,0 +1,180 @@
...
+
+# The core manifest declares these permissions and bindings, but
+# ec-bootstrap skips reapplying the core manifest once its RepoElement is
+# ACTIVE, so an already installed stand never seeds them. Seed them here so
+# the upgrade path reaches them too.
+#
+# The UUIDs are the ones the manifest declares: iam_permissions.name is
+# unique, so a row seeded under a different UUID collides with the manifest
+# on a fresh installation.
+SECRET_PERMISSIONS = [
```
akremenetsky: Could you explain why we need that?

exordos_core/gservice/service.py
slashburygin: ssh_key.  ok, in next pr i will delete this class

exordy (approved):
### exo-review: 1 low, nothing blocking · since ac8b12e: 0 fixed, 1 still open, 0 new

- `exordos_core/cmd/bootstrap.py:451` **[low]** _ensure_ua_config_current's docstring lists 3 reapply steps while the body performs 4.

#### Verify before merge

- _ensure_section against a real persisted /etc/exordos/core_agent.conf from an upgraded stand — copy one into tmp and call bootstrap._ensure_core_agent_config_current().

<details>
<summary>Coverage: 10 axes clean · 7 checks run · mode shipped</summary>

**Clean axes:** security — value exposure matches passwords; read stays project-scoped; compat — no target-resource or schema shape change in this range; compat — _ensure_section is a noop on the shipped template, test pins it; compat (prior) — 0004 in-branch, downgrade FK order right; reconciliation — config edits idempotent, repeated runs converge; tests — functional asserts flipped to the new contract; helper unit-tested; docs — no stale write-only claim left; secrets.md has no ru twin; scope — both commits trace to the stated intent; hygiene — commit format, English, no orphaned constants; prod-readiness — no TODO/WIP/history comments in the diff

**Checks run:**
- grep _CORE_AGENT_* repo-wide: no orphaned constant refs
- read RA packers.py: _skip_none=True, unset fields omitted
- grep dump_to_simple_view in restalchemy: not on the REST path
- grep write-only claims in docs/exordos_core/exordos: none left
- read core_agent.conf.j2: every declared key already matches
- read bootstrap.py:440-490: docstring list still at 3 steps
- axis-9 greps over the diff: clean

</details>

<sub>21 turns · 253s</sub>

migrations/0004-add-opaque-secrets-362f6d21.py
slashburygin: like in squashed migration.)
Remove this code

exordos_core/user_api/secret/api/controllers.py
slashburygin: ok, do not hide values

exordos_core/cmd/bootstrap.py
slashburygin: done

exordos/manifests/core.yaml.j2
slashburygin: remove uuids

exordy (approved):
### exo-review: 1 low, nothing blocking · since 3766816: 0 fixed, 1 still open, 0 new

- `exordos_core/cmd/bootstrap.py:451` **[low]** _ensure_ua_config_current's docstring still lists 3 reapply steps while the body performs 4.

#### Verify before merge

- That a core element upgrade really creates the four secret.secret.* permissions on an existing stand: upgrade core on a stand, then query iam_permissions.
- Dev stands that applied the earlier 0004 hold bindings at the now-unpinned UUIDs; check for duplicate owner rows in iam_binding_permissions.

Note: Head reverts the migration seeding, so @phantomii's change request on the upgrade path is again unaddressed by design.

<details>
<summary>Coverage: 8 axes clean · 5 checks run · mode shipped</summary>

**Clean axes:** compat — full_spec Secret_Create/_Get now matches Password's generated shape; compat — migration keeps only table DDL; downgrade drops index before table; security — permission UUIDs still pinned in the manifest, names unique; scope — both commits trace to the stated revert; hygiene — commit format, English, no orphaned constant or UUID refs left; reconciliation — no target-resource or agent-config change in this range; prod-readiness — axis-9 greps over the range: clean; tests — nothing referenced the removed seeding

**Checks run:**
- read pr-637.patch: whole range
- diff full_spec Secret vs Password schema: same generated shape
- grep removed perm/binding UUIDs + OWNER_ROLE_UUID: no dangling refs
- read migrations/0004 head: tables only
- read bootstrap.py:435-520, 804-866

</details>

<sub>10 turns · 154s</sub>

akremenetsky (approved):

PR #638, opened 2026-09-11
Install upgrade targets marked installed before registration

slashburygin:
## Summary

An element upgrade requested shortly after a repository sync silently did
nothing: the target version stayed `AVAILABLE` and the element kept running the
previous version forever. This is the upgrade twin of #630, which fixed the same
race for a fresh installation but left the upgrade case open.

The target of an upgrade carries the runtime element (`RepoElement.upgrade()`
transfers it), so `_require_installation` is false and `_require_upgrade` is
true. The create path only handled `_require_installation`, so when the builder
registered the target version it produced no `InstalledManifest` derivative and
pinned the status to `AVAILABLE`. The instance is marked as tracked right after
the resource creation, so the update path never ran either. The old version in
turn blocks in `_require_release`, waiting for the pending element's
`InstalledManifest` that is never created.

Observed on a stand as `core` staying on 0.2.27 after `exordos e e update core`:

```text
core 0.2.27  status ACTIVE     installation UNINSTALLED   <- old, waiting to be released
core 0.2.28  status AVAILABLE  installation INSTALLED     <- target, never installed
```

with no 0.2.28 element in EM at all.

## Changes

- Handle `_require_upgrade` alongside `_require_installation` in
  `create_instance_derivatives` and `can_create_instance_resource`
- Functional test reproducing the race: a new version appears through a
  repository refresh, the upgrade is requested, and the builder's only
  iteration for the target takes the create path
- Functional test for the ordinary upgrade of an already registered element
- Unit tests for both create hooks with an upgrade target

## Verification

- `.tox/py314-functional/bin/python -m pytest exordos_core/tests/unit/repo
  exordos_core/tests/functional/service exordos_core/tests/functional/restapi/repo`
  — 231 passed. The new functional test fails on master with
  `assert 'AVAILABLE' == 'IN_PROGRESS'`.
- `ruff check exordos_core` — clean; `ruff format --check` clean for the touched
  files (`cmd/bootstrap.py` and `tests/unit/cmd/test_bootstrap_ua_config.py` are
  unformatted on master already and are not touched here).
- Not verified on a stand: the `core_update_tests` job in `exordos_tests` is the
  end-to-end check for this and should be re-run against a build with this fix.

## Summary by Sourcery

Install upgrade targets through the create path when they are already marked for installation at registration time.

Bug Fixes:
- Ensure upgrade targets are installed when they are registered after an upgrade request, preventing them from remaining AVAILABLE indefinitely.

Tests:
- Add functional coverage for upgrades requested before and after target registration, plus unit coverage for upgrade-target creation hooks.

### DIFF
--- a/exordos_core/repo/builders/element.py
+++ b/exordos_core/repo/builders/element.py
@@ -433,10 +433,10 @@ class RepoElementBuilderService(
         called again on the next iteration.
         """
         # An element may already be marked as installed by the time the
-        # builder registers it: the installation request arrives between the
-        # repository sync and this iteration. Such an element is installed
-        # right away, so its dependencies must be resolvable first.
-        if self._require_installation(instance):
+        # builder registers it: the installation or upgrade request arrives
+        # between the repository sync and this iteration. Such an element is
+        # installed right away, so its dependencies must be resolvable first.
+        if self._require_installation(instance) or self._require_upgrade(instance):
             return self._dependencies_available(instance)
 
         return True
@@ -447,11 +447,11 @@ class RepoElementBuilderService(
         """Compute the derivative resources for a new instance.
 
         The hook is called only for new instances. An element that is already
-        installed when the builder registers it must be installed here: the
-        update path is never taken for it, because the instance is marked as
-        tracked right after the resource creation.
+        installed, or is an upgrade target, when the builder registers it must
+        be installed here: the update path is never taken for it, because the
+        instance is marked as tracked right after the resource creation.
         """
-        if self._require_installation(instance):
+        if self._require_installation(instance) or self._require_upgrade(instance):
             return [self._install_manifest(instance)]
 
         return ()
--- a/exordos_core/tests/functional/service/test_repo_builder.py
+++ b/exordos_core/tests/functional/service/test_repo_builder.py
@@ -96,6 +96,26 @@ def bootstrap_repo(manifests_dir, user_api):
 
 
 @pytest.fixture
+def bootstrap_repo_upgrade(manifests_dir, user_api):
+    """Create a Repository with two versions of the same element."""
+    _write_manifest(manifests_dir, "core", "1.0.0")
+    _write_manifest(manifests_dir, "core", "1.1.0")
+
+    repo = repo_builder.Repository(
+        name="test-repo-upgrade",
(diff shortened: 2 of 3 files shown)


exordy (commented):
### exo-review: 1 low, nothing blocking

Intent: Register an upgrade target through the create path so it installs instead of being pinned AVAILABLE.

#### Outside the diff

- `exordos_core/repo/builders/element.py:425` **[low]** The ERROR status of an unresolvable dependency closure is never persisted, so a failed upgrade target stalls silently. **Fix:** Persist the status in `_dependencies_available` (or have the caller save the instance when the gate fails). <details><summary>Why</summary>

`can_create_instance_resource` returning False skips `_actualize_new_instance`, the only place that saves the instance. The row keeps NEW, and the old version stays blocked in `_require_release` waiting for a derivative that is never created.

</details> 

#### Verify before merge

- End-to-end on a stand: re-run the `core_update_tests` job in exordos_tests against a build with this fix.
- That the old version actually releases after the target installs — the tests stop at the derivative; needs the agent filling the actual resource's `element`.

Note: PR description checks out against the diff: both hooks, both functional tests, the unit tests.

Note: test_upgrade_element_via_iteration passes on master too — regression coverage, not a reproduction, as stated.

<details>
<summary>Coverage: 8 axes clean · 4 checks run · mode shipped</summary>

**Clean axes:** security — builder-internal branch, no API, tenancy or input surface; compat — no migration, model, manifest or target-resource shape change; scope — 3 files, every hunk traces to the intent; reconciliation — per-instance try/except + savepoint; converges once the derivative exists; contracts — no signature or hook semantics changed, mirrors the update path; prod-readiness — no TODO/debug/history comments, conventional commit with the mechanism; tests — create path covered (red on master), update path gets a regression test; hygiene — English throughout, no generated artefacts, no dependency change

**Checks run:**
- grep TODO/FIXME/WIP over the patch — no hits
- grep history-style comments over the patch — no hits
- grep target-resource surface over the patch — no hits
- read builder.py _actualize_new_instance/_actualize_new_instances — derivatives persist

</details>

<sub>25 turns · 341s</sub>

akremenetsky (approved):

gmelikov (approved):

PR #641, opened 2026-09-14
Skip stored elements when creating a repository

slashburygin:
## Summary

A database repository builds its inventory from the elements already stored
for it. If an element was uploaded before the repository builder picked the
repository up, the create hook inserted that element again, hit
`repo_elements_repository_name_version_idx`, and rolled back. The repository
stayed `NEW` and failed the same way on every builder iteration.

## Changes

- `post_create_instance_resource` skips inventory elements already stored for
  the repository, the same comparison `_refresh_repository` does
- The "Saved N elements" log line counts only inserted elements
- New functional test: upload to a database repository before the first
  builder iteration, then expect `ACTIVE` with only the uploaded element

## Verification

- New test failed before the fix with the production `UniqueViolation`, passes after
- `pytest exordos_core/tests/unit`: 281 passed
- `pytest` on functional tests touching repositories (local Postgres): 96 passed, 3 skipped
- `ruff check` and `ruff format --check` on changed files: clean
- `mypy -p exordos_core`: 603 errors, identical count on master (no new errors)
- Full functional suite: not run

## Summary by Sourcery

Skip already stored elements when initializing repository inventory to avoid duplicate entries and repeated creation failures.

Bug Fixes:
- Prevent repository creation from re-inserting elements that were uploaded before the first builder iteration, allowing the repository to activate successfully.

Enhancements:
- Ensure element-save logging reports only newly inserted inventory elements.

Tests:
- Add functional coverage for preserving pre-existing uploads during database repository activation.


exordy (approved):
### exo-review: LGTM

Intent: Skip inventory elements already stored when a repository is created, so a pre-upload doesn't hit the unique index and strand the repo in NEW.

#### Verify before merge

- CI's ruff/pytest/mypy runs per PR description, not re-run here per instructions

Note: Sourcery's open comment at repository.py:101 is correct (latest stays False for skipped elements); same known gap as TODO at controllers.py:85. Follow-up, not a blocker.

<details>
<summary>Coverage: 8 axes clean · 0 checks run · mode pre-prod</summary>

**Clean axes:** security — query scoped by instance.uuid, matches _refresh_repository's own filter; diff scope — touches only the create hook and its test, matches commit body; bugfix self-explains — new test reproduces the UniqueViolation pre-fix, passes post-fix; commit claims — log now counts saved_count not i+1; mirrors _refresh_repository keying; minimal fix — skips already-stored (name, version) pairs, nothing more; idempotency — hook is now safe to re-run without re-inserting; hygiene — no TODO/debug/dead code added; comment explains why, not what; compat — no target-resource/migration/API shape touched, N/A

</details>

<sub>20 turns · 109s</sub>

exordos_core/repo/builders/repository.py
```diff
@@ -85,19 +85,32 @@ def post_create_instance_resource(
         """
         super().post_create_instance_resource(instance, resource, derivatives)
 
+        # Elements may already be stored, for example uploaded to a database
+        # repository whose inventory is built from these very rows.
+        stored_elements = {
```
akremenetsky, line 90: I don’t really like this approach. It looks more like patching a symptom than addressing the underlying conceptual issue.

I think we should reject uploads until repository provisioning is complete. The repository status should be checked in `Repository.upload()`, and uploads should only be allowed when the repository is `ACTIVE`. This would establish a clear lifecycle invariant and eliminate the race between the initial provisioning and an upload, making the stored_elements workaround unnecessary.

slashburygin: ok, only check active status for repo, but firstly we need to merge this pr
https://github.com/exordos/exordos/pull/400

exordy (approved):
### exo-review: LGTM

#### Verify before merge

- Merge exordos/exordos#400 before this one. Otherwise the first `exordos em elements install` hits a NEW repository and is rejected.

Note: Released CLIs without #400 fail the first install against an upgraded core (a rerun works). Release the CLI together with or before core.

<details>
<summary>Coverage: 10 axes clean · 4 checks run · mode shipped</summary>

**Clean axes:** security — status gate only; the existing `_enforce("upload")` still runs first; diff scope — gate, revert, test helpers, docs; all listed in the PR; bugfix self-explains — commit body gives the mechanism; new test covers rejection; commit claims — builder hook back to the master version; helpers activate repos; minimal fix — one status check; NEW→ACTIVE is one-way, SDK update keeps ACTIVE; docs — repo-proxy.md and .ru.md both updated; compat — no target-resource, migration or schema change; API gains a 4xx precondition; reconciliation — refresh keys on stored elements; no duplicate insert remains; prod-readiness — no TODO/debug markers or process comments added; hygiene — commit format ok; no new files, generated artefacts or deps

**Checks run:**
- grep patch for WIP/TODO markers: none
- grep SDK builder update path: no IN_PROGRESS on update, ACTIVE persists
- grep upload callers: only user_api controller; test callers use ACTIVE repos
- read exordos#400: waits for ACTIVE, fails fast on ERROR/DISABLED

</details>

<sub>10 turns · 56s</sub>

PR #643, opened 2026-09-17
Add MCP server endpoint to User API

slashburygin:
## Summary

Adds a Model Context Protocol endpoint to the User API at `/v1/mcp` (`/api/core/v1/mcp` on an installation), so AI agents can manage the platform with the same credentials and permissions as a direct API client.

## Changes

- `exordos_core/user_api/api/mcp.py`: a stateless MCP endpoint over Streamable HTTP. Each request is one JSON-RPC POST with one JSON response. There are no sessions, no server-sent events and no batching. It is written by hand because the official SDK only supports asyncio, and this API runs on bjoern/WSGI.
- Tools:
  - `list_endpoints`: lists endpoints, with an optional search filter.
  - `describe_endpoint`: returns one OpenAPI operation with every `$ref` resolved.
  - `call_api`: calls an endpoint and returns the HTTP status and body.
- `call_api` sends the call through the whole middleware stack with the caller's `Authorization` and `X-OTP` headers and the forwarding headers. IAM permissions and security rules therefore apply exactly as they do to a direct call. For this reason the MCP middleware is the outermost layer in `app.py`: the IAM middleware refuses to open a session inside another one.
- A request without a Bearer header gets `401` with `WWW-Authenticate: Bearer`. Only `call_api` checks that the token is valid. The list and describe tools return the OpenAPI document, which is already public in the docs.
- The OpenAPI document is built on first use in each worker, which takes about 5 seconds.
- `server.json` (MCP Registry format, `version: latest`) describes the installation and direct endpoints, with the token as a secret header; it can be imported into MCP Inspector. It passes validation against the `2025-12-11` registry schema.
- Docs: `docs/usage/mcp.md` and `docs/usage/mcp.ru.md`, a nav entry and its Russian label, and a mention in the README.

## Verification

- `pytest exordos_core/tests/unit/user_api/api`: 22 passed
- `pytest exordos_core/tests/functional -n 10`: 570 passed, 3 skipped. This includes the new MCP tests, one of which checks that a user without rights gets 403 through `call_api`.
- `pytest exordos_core/tests/unit -n 10`: all pass except the 2 known failures in `boot_api/test_models.py::TestMachineNetbootRestore`, which also fail on a clean `master`.
- `ruff check` and `ruff format --check`: clean. Markdownlint on the changed Markdown files: clean.
- Not run: tox itself and mypy (both broken in the local environment), and the docs site build.

## Summary by Sourcery

Enable AI agents to manage Exordos through a dedicated, permission-preserving MCP service alongside the User API.

New Features:
- Add a stateless MCP server endpoint that lets AI agents discover and invoke User API operations through JSON-RPC over Streamable HTTP.
- Provide generated resource-specific tools and generic endpoint discovery, description, and API-call tools based on the User API OpenAPI specification.
- Expose the MCP service through the installation load balancer and package it as a managed systemd service with configurable binding and User API settings.

Enhancements:
- Preserve caller authentication, security proofs, forwarding context, IAM permissions, and security rules when MCP tools invoke the User API.
- Improve MCP request validation, error reporting, UUID handling, and protection against unsupported paths and methods.

Build:
- Add the `ec-mcp-api` entry point and update the restalchemy dependency floor required by the MCP/OpenAPI integration.

Deployment:
- Configure, install, enable, and route the MCP service on port 11014 in deployed installations.

Documentation:
- Add English and Russian MCP usage documentation covering authentication, tools, configuration, and client integrations.
- Register the MCP server in the documentation navigation, README, and MCP Registry metadata.

Tests:
- Add unit and functional coverage for MCP transport behavior, protocol handling, tool generation, API forwarding, validation, and permission enforcement.


exordos_core/user_api/api/mcp.py
```diff
@@ -0,0 +1,312 @@
...
+FORWARDED_HEADERS = (
+    "HTTP_AUTHORIZATION",
+    "HTTP_X_OTP",
+    "HTTP_X_FORWARDED_FOR",
+    "HTTP_X_FORWARDED_HOST",
+    "HTTP_X_FORWARDED_PORT",
+    "HTTP_X_FORWARDED_PREFIX",
+    "HTTP_X_FORWARDED_PROTO",
+    "HTTP_X_REAL_IP",
+    "REMOTE_ADDR",
```
exordy, line 64: **[medium] `HTTP_HOST` is not forwarded, so replayed calls see host `localhost` wherever the API builds an absolute URL from `req.host_url`.**

`Request.blank` fills `HTTP_HOST` with `localhost:80`, and the copied `X-Forwarded-*` are read only by gcl_iam's context — exordos_core's own call sites use `req.host_url` directly (iam/api/controllers.py:113, :804, :922).

</details>

```suggestion
    "HTTP_X_REAL_IP",
    "HTTP_HOST",
    "REMOTE_ADDR",
```

exordos_core/user_api/api/mcp.py
```diff
@@ -0,0 +1,312 @@
...
+            return _result(msg_id, {"tools": TOOLS})
+        if method == "tools/call":
+            handler = getattr(self, f"_tool_{params.get('name')}", None)
+            if handler is None:
+                return _error(
+                    msg_id, INVALID_PARAMS, f"Unknown tool: {params.get('name')}"
+                )
+            try:
+                text, is_error = handler(req, **(params.get("arguments") or {}))
+            except (ToolError, TypeError) as e:
```
exordy: **[medium] Only `ToolError` and `TypeError` are caught, and this middleware sits outside `ErrorsHandlerMiddleware`, so other tool failures become a bare 500.**

Nothing validates argument types against `inputSchema`: `search=null` hits `None.lower()` (line 241) and a non-string `path` hits `.startswith` (line 263), both AttributeError. Being the outermost layer, nothing below converts it into a response.

</details>

```suggestion
            except Exception as e:
```

exordos_core/user_api/api/mcp.py
```diff
@@ -0,0 +1,312 @@
...
+        self._specification = None
+
+    @dec.wsgify
+    def __call__(self, req):
+        if req.path_info.rstrip("/") != MCP_PATH:
+            return req.get_response(self.application)
+        if req.method != "POST":
+            # No server-initiated stream (GET) and no session to end (DELETE).
+            return webob.Response(status=405, headers={"Allow": "POST"})
+        if not req.authorization:
```
exordy: **[low] The gate checks that an `Authorization` header exists, not that it is a Bearer token, so `Authorization: x` reaches `list_endpoints` and the 5-second spec build.**

`req.authorization` is truthy for any non-empty header value. The docs state an IAM access token is required; in practice only `call_api` is validated, by the stack below.

</details>

```suggestion
        if not req.authorization or req.authorization[0].lower() != "bearer":
```

exordy (commented):
### exo-review: 2 medium, 1 low, nothing blocking

Intent: Add a stateless MCP endpoint to the User API with three OpenAPI-driven tools, plus docs and tests.

#### Verify before merge

- The spec build inside a live worker (DB up, element registry loaded): start ec-user-api and call list_endpoints — functional tests only cover call_api.
- Whether the ~5 s build stalls the worker's other requests: issue a second request while the first list_endpoints runs.

<details>
<summary>Coverage: 8 axes clean · 3 checks run · mode shipped</summary>

**Clean axes:** security — call_api replays the full stack; IAM and rules still decide; tenancy — no query or filter touched; scope — every hunk is the endpoint, its docs or its tests; compat — additive path; other requests only pass through the new layer; reconciliation — no surface; prod-readiness — no WIP markers, both doc languages, Apache headers; tests — transport, protocol, tools, and a 403 through call_api; hygiene — conventional subject, no generated artefacts committed

**Checks run:**
- grep over pr-643.patch for TODO/WIP/process comments: none
- grep for X-Forwarded/host_url in core + siblings: only gcl_iam reads them
- read app.py, common/openapi.py, iam/api/controllers.py host_url sites

</details>

<sub>23 turns · 392s</sub>

exordos_core/user_api/api/mcp.py
```diff
@@ -0,0 +1,312 @@
...
+                    "instructions": INSTRUCTIONS,
+                },
+            )
+        if method == "ping":
+            return _result(msg_id, {})
+        if method == "tools/list":
+            return _result(msg_id, {"tools": TOOLS})
+        if method == "tools/call":
+            handler = getattr(self, f"_tool_{params.get('name')}", None)
+            if handler is None:
```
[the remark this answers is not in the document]
slashburygin: Fixed in 073f2d3: non-object `params`/`arguments` now get `INVALID_PARAMS`, and tool arguments are checked against each tool's `inputSchema` (required, unknown, type), so wrong types become tool errors instead of exceptions.

exordos_core/user_api/api/mcp.py
slashburygin: Fixed in 073f2d3: `HTTP_HOST` and `wsgi.url_scheme` are forwarded; the unit test now checks `host_url` on the replayed request.

exordos_core/user_api/api/mcp.py
slashburygin: Fixed in 073f2d3 by validating arguments against the input schema before dispatch rather than catching every exception, so real bugs still surface. Only `ToolError` is caught now.

exordos_core/user_api/api/mcp.py
slashburygin: Fixed in 073f2d3: only the `Bearer` scheme passes; anything else gets 401 with `WWW-Authenticate: Bearer`.

slashburygin:
Re the exo-review verify items:
- Spec build in a live worker: covered by the new functional test `test_discovery_in_a_running_service` (list_endpoints and describe_endpoint against the running service, then an ordinary call to confirm it still responds normally).
- Stall: yes, the ~5 s first build blocks that worker's other requests, once per worker, because each worker handles one request at a time. Building at worker start would slow every worker's startup, including workers that never serve MCP, so it stays on first use.

exordos_core/user_api/api/mcp.py
```diff
@@ -0,0 +1,344 @@
...
+
+# The JSON types the tool input schemas above use.
+JSON_TYPES = {"string": str, "object": dict}
+
+
+class ToolError(Exception):
+    """A tool failed in a way the model should read and react to."""
+
+
+class McpMiddleware:
```
akremenetsky, line 160: Would we like to put in the RA?

exordos_core/user_api/api/app.py
```diff
@@ -80,5 +81,7 @@ def build_wsgi_application(context_storage, iam_engine_driver):
             ),
             errors_mw.ErrorsHandlerMiddleware,
             logging_mw.LoggingMiddleware,
+            # Outermost: its tool calls re-enter the whole stack above.
+            mcp.McpMiddleware,
```
akremenetsky, line 85: Could we implement MCP as a small separate service rather than as an outer middleware? The service could expose the MCP endpoint on its own internal port and call the standard User API over HTTP, forwarding the caller’s Bearer token and OTP header. This would keep authentication and authorization entirely within the existing User API, avoid re-entering the WSGI middleware stack, and provide a cleaner separation between the MCP transport and the application API. It would also allow MCP to evolve and scale independently without adding protocol-specific behavior to the User API middleware chain.

exordos_core/user_api/api/mcp.py
```diff
@@ -0,0 +1,344 @@
...
+                        "name": "exordos-core-user-api",
+                        "version": version.version_info,
+                    },
+                    "instructions": INSTRUCTIONS,
+                },
+            )
+        if method == "ping":
+            return _result(msg_id, {})
+        if method == "tools/list":
+            return _result(msg_id, {"tools": TOOLS})
```
akremenetsky, line 220: Could we expose a curated set of domain-specific tools while keeping list_endpoints, describe_endpoint, and call_api as a fallback for unsupported or uncommon operations? Tools such as create_node and update_node, with explicit input and output schemas, would be easier and more reliable for models than requiring them to discover and interpret raw HTTP endpoints for every common task. The generic endpoint tools would still preserve complete API coverage without requiring every operation to be represented immediately. As an initial set, could you please add domain-specific tools for nodes, node sets, configs, values, secrets, elements, users, projects, organizations, and roles?

exordos/manifests/core.yaml.j2
```diff
@@ -1012,6 +1020,30 @@ resources:
           - kind: rewrite_url
             regex: "^/api/core/(.*)"
             replacement: "/$1"
+    core_lb_mcp_route:
+      project_id: "12345678-c625-4fee-81d5-f691897b8142"
+      parent: $core.network.lb.$core_lb.vhosts.$core_lb_core_http:uuid
+      condition:
+        kind: prefix
+        value: /api/core/v1/mcp
```
akremenetsky, line 1028: Actually I'm sure we need to place `/mcp` under the user API. Seems it's something separated.

exordos_core/mcp_api/api/mcp.py
```diff
@@ -0,0 +1,548 @@
...
+                    "serverInfo": {
+                        "name": "exordos-core-user-api",
+                        "version": version.version_info,
+                    },
+                    "instructions": INSTRUCTIONS,
+                },
+            )
+        if method == "ping":
+            return _result(msg_id, {})
+        if method == "tools/list":
```
akremenetsky, line 285: The generated schemas currently include a lot of OpenAPI metadata that is useful for API documentation but expensive and distracting in an LLM tool definition. For example, the generated create_node definition is about 2.1 KB, while removing only example, title, maxLength, and pattern reduces it to about 1.2 KB.

Could we apply a recursive compacting pass before exposing these schemas through MCP? I suggest retaining fields that help the model construct a valid request—such as type, required, enum, format, nullable, useful defaults, and numeric limits—while dropping documentation-oriented metadata such as repeated examples, titles, length limits, and regex patterns.

The largest remaining schemas come from complex nested fields such as disk_spec. Could we provide a small curated override for these fields that describes the preferred/common case and direct uncommon cases to call_api? For example:
```
{
  "type": "object",
  "description": "Root disk specification. Use call_api for uncommon multi-disk configurations.",
  "properties": {
    "kind": {
      "type": "string",
      "enum": ["root_disk"]
    },
    "image": {
      "type": "string",
      "description": "Root disk image."
    },
    "size": {
      "type": "integer",
      "minimum": 1,
      "default": 10,
      "description": "Root disk size in GiB."
    }
  },
  "required": ["kind", "image"]
}
```
This would preserve OpenAPI as the source of the top-level HTTP contract while making the curated tools substantially smaller and easier for models to use.

PR #645, opened 2026-09-18
Move SSH keys to their own builder service

slashburygin:
## Summary

SSH keys were the last secret kind still driven by the hand-rolled
`SecretServiceBuilder` loop, while secrets, passwords and certificates had
already moved to the universal builder. This ports them over: `SSHKeyBuilder`
runs on `UniversalBuilderService` and the per node "slave" resources become
SDK derivatives of the master key resource.

With SSH keys gone, `SecretServiceBuilder` held only generic secret helpers
that nothing called and an empty iteration, so `exordos_core/secret/service.py`
is removed along with the `SSHKey` model helpers that only it used.

## Changes

- `SSHHostKey` builder model: kind `ssh_key_target`, scheduled to the node's
  agent through `agent_uuid`, initial resource status `IN_PROGRESS`. Target
  fields are declared explicitly so `agent_uuid` stays out of the resource
  value and the hash the data plane computes is unchanged.
- `SSHKey` becomes an `InstanceWithDerivativesMixin` mapping
  `ssh_key_target` to `SSHHostKey`.
- `SSHKeyBuilder` keeps the behavior of the old loop: delete a key whose
  owners are gone, wait for at least one node, one host key per target node
  (uuid5-derived as before), back to `IN_PROGRESS` on update, and roll the per
  node statuses up into the status of the key.
- `SecretServiceBuilder` and `exordos_core/secret/service.py` removed;
  `SSHKeyBuilder` registered in `GeneralService` in its place.

Two behavior differences follow from the SDK taking over:

1. Updating a key no longer deletes and recreates every per node resource.
   The SDK diffs the derivative set, so unchanged nodes keep their resource
   and a node added to a set gets one without touching the others.
2. The status roll-up now sees all derivatives, not only outdated ones. A node
   that has not reported yet counts as `IN_PROGRESS`, so a key no longer
   reports `ACTIVE` while some node is still missing it.

## Verification

- Functional suite: 567 passed, 3 skipped
  (`pytest -n 10 exordos_core/tests/functional`). The five existing SSH key
  tests pass with their assertions untouched - only the class under test was
  repointed at `SSHKeyBuilder`.
- Unit suite: 285 passed, 2 failed
  (`pytest -n 10 exordos_core/tests/unit`). The two failures are
  `boot_api/test_models.py::TestMachineNetbootRestore`, which fail the same
  way on master and are unrelated to this change.
- `ruff check .` clean; `ruff format --check` clean for every file touched
  here (it flags two untouched files that already drift on master).
- Checked directly that the host key resource hash is byte-identical to the
  one the old `to_host_resource` produced, so the data plane sees exactly
  what it saw before.
- tox was not used: the tox envs are broken in this checkout, so the gates
  above ran through the env interpreters directly.

## Summary by Sourcery

Move SSH key delivery to the universal builder and represent each node-specific key as a derivative of the master resource.

New Features:
- Deliver SSH keys through the universal builder as per-node derivative resources.

Bug Fixes:
- Ensure SSH key status remains in progress until every target node has reported the expected key.
- Remove SSH keys whose owners no longer exist while retaining the existing delivery behavior.

Enhancements:
- Preserve unchanged per-node resources during key updates and create or remove derivatives only as target nodes change.
- Remove the obsolete hand-rolled secret service and SSH key helper methods.

Tests:
- Update functional SSH key coverage for the universal builder and add scenarios for stable updates, missing owners, and multi-node status roll-up.


exordos_core/secret/builders/service.py
```diff
@@ -192,3 +228,107 @@ def actualize_outdated_instance(
...
+            for node in instance.target_nodes()
+        ]
+
+    def post_update_instance_resource(
+        self,
+        instance: SSHKey,
+        resource: ua_models.TargetResource,
+        derivatives: tp.Collection[ua_models.TargetResource] = (),
+    ) -> None:
+        instance.status = sc.SecretStatus.IN_PROGRESS.value
```
exordy: **[medium] An update that touches no per-node field pins the key in IN_PROGRESS permanently.**

**Fix:** Roll the per-node statuses up here instead of forcing IN_PROGRESS, or force it only when a derivative hash actually changed.

The only path back to ACTIVE is the roll-up, driven by ua_outdated_resources_view (target.full_hash != actual.full_hash). With derivative values unchanged, _actualize_derivative_target_resources returns early, no derivative row is touched, and no such row ever appears.

</details>

exordos_core/tests/functional/service/test_secrets.py
```diff
@@ -20,15 +20,15 @@
...
 from gcl_sdk.agents.universal.dm import models as ua_models
 
-from exordos_core.secret import service
+from exordos_core.secret.builders import service
 from exordos_core.secret.dm import models
 from exordos_core.tests.functional import stubs
 
 
-class TestSecretsServiceBuilder:
+class TestSSHKeyBuilder:
```
exordy, line 33: **[low] Both behavior changes the PR claims are untested: all five tests use a single node.**

**Fix:** Add a node_set case with two nodes where only one reports ACTIVE.

With one derivative, all()/any() over one status cannot distinguish the new roll-up from the old, and the derivative-set diff has nothing to diff. A node_set key with two nodes, one reporting, is the case the description calls out.

</details>

exordy (commented):
### exo-review: 1 medium, 2 low, nothing blocking

Intent: Port SSH keys from the hand-rolled SecretServiceBuilder loop onto the SDK universal builder.

#### Outside the diff

- `exordos_core/tests/functional/service/test_secrets_builder.py:84` **[low]** TestSecretServiceBuilder now names a class this PR deleted. **Fix:** Rename to match what it exercises (the universal secret builders). <details><summary>Why</summary>

The name was already loose, but it is this commit that removes the SecretServiceBuilder it refers to, leaving the file pointing at nothing.

</details> 

#### Verify before merge

- Finding 1 on a stand: PUT only `description` on a delivered key, then watch its status across a few builder iterations.
- Multi-node roll-up: add a node_set fixture and run `pytest exordos_core/tests/functional/service/test_secrets.py` on a clean database.

Note: The description's "a node added to a set gets one" holds on update only; the outdated path reuses existing derivatives.

<details>

**Clean axes:** compat — resource value, hash, kind and uuid5 all byte-identical; no migration owed; security — no API, tenancy, permission or log surface touched; scope — every hunk traces to the port; the two FIXMEs move verbatim; reconciliation — delete, update and roll-up paths settle, except finding 1; prod-readiness — no WIP markers, debug output or process comments; hygiene — English throughout, no generated artefacts, no dependency change

**Checks run:**
- read pr-645.patch in full — 790 lines
- hygiene greps over the patch — clean
- grep for refs to the deleted service/helpers — none left
- traced gcl_sdk builder.py create/update/outdate/delete paths
- checked ua_outdated_resources_view SQL in gcl_sdk migration 0004

</details>

<sub>44 turns · 522s</sub>

exordos_core/secret/builders/service.py
```diff
@@ -192,3 +228,107 @@ def actualize_outdated_instance(
...
+    def actualize_outdated_instance_derivatives(
+        self,
+        instance: SSHKey,
+        derivative_pairs: tp.Collection[tp.Tuple[SSHHostKey, tp.Optional[SSHHostKey]]],
+    ) -> tp.Collection[SSHHostKey]:
+        """Roll the per node statuses up into the status of the key."""
+        statuses = self._host_key_statuses([t for t, _ in derivative_pairs])
+
+        if all(s == sc.SecretStatus.ACTIVE for s in statuses):
+            instance.status = sc.SecretStatus.ACTIVE.value
```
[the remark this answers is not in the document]
slashburygin: Not reachable, so left as is.

`actualize_outdated_instance_derivatives` is only called from `_actualize_outdated_instance_derivatives` / `..._all_derivatives`, and both return early when `changed_derivatives` is empty. `all_derivatives` comes from `_get_resources_by_masters`, which is a superset of the changed ones for the same master, so `derivative_pairs` is never empty here.

The empty-target case you describe is real, but it takes the update path, and there it left the key in `IN_PROGRESS`, not `ACTIVE`. Fixed in 0043458: `can_update_instance_resource` now revalidates the owners, so a key whose target set is emptied is deleted (owners gone) or waits for a node, the same way as on creation.

exordos_core/secret/builders/service.py
```diff
@@ -192,3 +228,107 @@ def actualize_outdated_instance(
...
+        iter_min_period: int = 1,
+        iter_pause: float = 0.1,
+    ) -> None:
+        super().__init__(
+            instance_model=SSHKey,
+            iter_min_period=iter_min_period,
+            iter_pause=iter_pause,
+        )
+
+    def can_create_instance_resource(self, instance: SSHKey) -> bool:
```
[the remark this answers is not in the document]
slashburygin: Confirmed, fixed in 0043458.

The owner check moved to `_can_deliver_instance` and is now called from `can_update_instance_resource` as well, so a key retargeted to a node that does not exist is deleted exactly as on creation. Without it the update did not just leave the key in `IN_PROGRESS`: scheduling the derivative to the missing agent raised, and the SDK swallowed it as `Error updating instance(ssh_key) resource ...` on every iteration.

Covered by `test_update_ssh_keys_fake_node`.

exordos_core/secret/builders/service.py
slashburygin: Confirmed, fixed in 0043458.

`name` and `description` are master target fields and are writable, so `PUT {"description": ...}` produced exactly this: the derivative set is unchanged, `_actualize_derivative_target_resources` returns early, no derivative row ever becomes outdated and the roll-up is never called again.

The roll-up is now extracted into `_roll_up_status` and runs in `post_update_instance_resource` too. At that point the derivative target resources already carry the new value while the actual ones still hold the old one, so a real change still lands on `IN_PROGRESS` and a no-op update goes straight back to `ACTIVE`.

Covered by `test_update_ssh_keys_without_node_changes`.

exordos_core/tests/functional/service/test_secrets.py
slashburygin: Fair, added in 0043458.

`test_node_set_ssh_key_waits_for_every_node` builds a node set of two nodes, lets only the first report the key back and asserts the key stays `IN_PROGRESS`, then `ACTIVE` once the second reports. That test passes on the previous commit as well, which is the point: it pins the roll-up behavior the description claims rather than a fix.

The other two new tests cover the update paths from the findings above, and both fail without 0043458.

exordos_core/secret/builders/service.py
```diff
@@ -192,3 +228,132 @@ def actualize_outdated_instance(
...
+        self,
+        instance: SSHKey,
+        resource: ua_models.TargetResource,
+        derivatives: tp.Collection[ua_models.TargetResource] = (),
+    ) -> None:
+        # An update does not have to touch a per node field, `name` for
+        # instance. Then no derivative resource changes, so nothing becomes
+        # outdated and the roll-up below is never called again. The status
+        # has to be rolled up here too, otherwise such a key would stay in
+        # the status forced here forever.
```
exordy, line 299: **[low] The comment describes the previous implementation rather than this one.**

"the roll-up below" and "the status forced here" refer to code that is no longer in this method; nothing is forced here now, so a later reader cannot resolve either phrase.

</details>

```suggestion
        # An update does not have to touch a per node field, `name` for
        # instance. Then no derivative resource changes, nothing becomes
        # outdated and `actualize_outdated_instance_derivatives` is never
        # called, so the status is rolled up here as well.
```

exordy (approved):
### exo-review: 2 low, nothing blocking · since 42cc79f: 2 fixed, 1 still open, 1 new

- `exordos_core/tests/functional/service/test_secrets_builder.py:84` **[low]** TestSecretServiceBuilder names the class this PR deletes.

Note: PR description still says "back to IN_PROGRESS on update"; 0043458 rolls the status up instead.

<details>
<summary>Coverage: 8 axes clean · 5 checks run · mode shipped</summary>

**Clean axes:** reconciliation — update, delete-on-missing-owner and roll-up all settle; status roll-up — actual host resources only ever report ACTIVE, so no dead branch bites; tests — the two update tests fail without 0043458; roll-up pinned over two nodes; compat — carried forward: value, hash, kind and uuid5 unchanged, no migration owed; security — no API, tenancy, permission or log surface touched; scope — both hunks trace to the two review threads; prod-readiness — no WIP markers, debug output or dead flags; hygiene — English throughout, no generated artefacts, no dependency change

**Checks run:**
- read the 274-line incremental diff and the head of SSHKeyBuilder
- traced SDK _actualize_updated_instance / _actualize_outdated_*_derivatives
- checked fetch_all_derivatives_on_outdate default (True) and update_instance_derivatives
- hygiene greps over the incremental diff — clean
- checked conftest cleanup_test_entities covers Node/NodeSet the new test inserts

</details>

<sub>25 turns · 323s</sub>

exordos_core/secret/builders/service.py
```diff
@@ -192,3 +228,132 @@ def actualize_outdated_instance(
...
+        # FIXME(akremenetsky): Only nodes as owners are supported for now.
+        # It will be updated when sets appear.
+
+        # FIXME(akremenetsky): Seems the key may be deleted since its
+        # owners are absent. May be it will be better to control this
+        # behavior via an additional option in the target model but for
+        # now just delete this config.
+        if not instance.target.are_owners_alive():
+            LOG.error("SSH key %s has no owners, delete it.", instance.uuid)
+            instance.delete()
```
akremenetsky, line 268: It's pretty strange to perform some destructive operations in a predicate method.

exordos_core/secret/builders/service.py
```diff
@@ -192,3 +228,132 @@ def actualize_outdated_instance(
...
+        # has to be rolled up here too, otherwise such a key would stay in
+        # the status forced here forever.
+        self._roll_up_status(instance, self.create_instance_derivatives(instance))
+
+    def actualize_outdated_instance_derivatives(
+        self,
+        instance: SSHKey,
+        derivative_pairs: tp.Collection[tp.Tuple[SSHHostKey, tp.Optional[SSHHostKey]]],
+    ) -> tp.Collection[SSHHostKey]:
+        host_keys = tuple(t for t, _ in derivative_pairs)
```
akremenetsky, line 306: Why we would like to get `target` resources. Then fetch `actual` resources:

```python
actual_resources = {
            r.uuid: r
            for r in ua_models.Resource.objects.get_all(
                filters={
                    "uuid": dm_filters.In([str(k.uuid) for k in host_keys]),
                    "kind": dm_filters.EQ(sc.SSH_KEY_TARGET_KIND),
                }
            )
        }
```

If we already have the `actual` resources:
```
host_keys = tuple(a for _, a in derivative_pairs)
```

Also you need to compare count of target and actual resources.

exordos_core/secret/builders/service.py
slashburygin: Fair, applied. Took your wording as is.

exordos_core/secret/builders/service.py
slashburygin: Carried over as is from `_actualize_new_ssh_key` (`git show 42cc79f^:exordos_core/secret/service.py`), FIXME included. Same pattern is already in the tree: `_can_create_machine` -> `_reschedule_machine` -> `machine.delete()` (`compute/builders/pool.py:172-178`).

The SDK has no pre-flight hook that may delete, so the alternative is a separate sweep over owner-less keys. Happy to do that here if you prefer.

exordos_core/secret/builders/service.py
slashburygin: `a` is not the actual resource, it is an `SSHHostKey` rebuilt from it - the SDK converts both sides before the hook (`gcl_sdk/agents/universal/services/builder.py:749-753`). It carries no `status` and no `full_hash`, which is what the roll-up reads, and it is `None` for a node that has not reported.

The count check is already there, per key: no actual resource -> `resource is None` -> `IN_PROGRESS`. And `_roll_up_status` is called from `post_update_instance_resource` too, which gets no pairs at all.

